PR #1, opened 2017-05-15
Add support for creating FixedDecimals from strings

omus:
Allows for precise creation of FixedDecimal values without using a floating point.

Continuation of https://github.com/JuliaFinance/Currencies.jl/pull/77

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -23,7 +23,7 @@
 
 module FixedPointDecimals
 
-export FixedDecimal
+export FixedDecimal, RoundThrows
 
 using Compat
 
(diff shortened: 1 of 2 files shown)


TotalVerb:
This is looking good, thanks. However, I think we need to figure out something to do in the case that the string has too much precision. Options are:

- throw InexactError?
- round?
- truncate? (what's being done right now)

With `Floatxx` types, rounding is done instead of `InexactError`. But `FixedDecimal` may have greater appetite for throwing an error here. What do you think?

TotalVerb:
(I added a few tests, some of them skipped while we decide what best to do on extra precision.)

omus:
I think rounding as the default is reasonable as that is what everything else does currently. I may add in a keyword argument to allow the user to select truncation if they prefer.

I like the concept of raising an exception if the string precision is greater than the precision of the FixedDecimal. At the moment however I'm not sure when a user would want this behaviour.

omus:
Supporting rounding resulted in me heavily refactoring the existing parse code. I ended up using `RoundingMode` to allow the user to specify the behaviour they want. Users can use `RoundNearest` (the default), `RoundToZero` (truncate), or the new rounding mode: `RoundThrows`. The new mode will raise an `InexactError` if the provided string cannot be precisely represented in the specified `FixedDecimal` type.

PR #2, opened 2017-05-16
Turn back on precompilation

TotalVerb:
Precompilation was inherited from Currencies originally, but was lost in the transfer. This restores it.


omus:
Thanks. Good catch.

PR #6, opened 2017-05-17
Better support for dividing by integers

omus:
Previously division would work by promotion but in some cases the integer provided would be outside of the range supported by the FixedDecimal:

```julia
julia> FD{Int8,1}(2) / FD{Int8,1}(10)
FixedDecimal{Int8,1}(0.2)

julia> FD{Int8,1}(2) / FD{Int8,1}(20)
ERROR: InexactError()
 in convert at /Users/omus/.julia/v0.5/FixedPointDecimals/src/FixedPointDecimals.jl:207 [inlined]
 in FixedPointDecimals.FixedDecimal{Int8,1}(::Int64) at ./sysimg.jl:53

julia> FD{Int8,1}(2) / 20  # With this PR
FixedDecimal{Int8,1}(0.1)
```

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -34,6 +34,8 @@ import Base: reinterpret, zero, one, abs, sign, ==, <, <=, +, -, /, *, div,
              typemin, typemax, realmin, realmax, print, show, string, convert,
              promote_rule, min, max, trunc, round, floor, ceil, eps, float, widemul
 
+import Base.Checked: checked_mul
+
 const IEEEFloat = Union{Float16, Float32, Float64}
 
 for fn in [:trunc, :floor, :ceil]
@@ -146,7 +148,7 @@ end
 # these functions are needed to avoid InexactError when converting from the integer type
 function /{T, f}(x::Integer, y::FD{T, f})
     powt = T(10)^f
-    xi = widemul(x, powt)
+    xi = checked_mul(x, powt)
     yi = y.i
     quotient, remainder = divrem(xi, yi)
     reinterpret(FD{T, f}, quotient * powt + round(T, remainder / yi * powt))
@@ -155,7 +157,7 @@ end
 function /{T, f}(x::FD{T, f}, y::Integer)
     powt = T(10)^f
     xi = x.i
-    yi = widemul(y, powt)
+    yi = checked_mul(y, powt)
     quotient, remainder = divrem(xi, yi)
     reinterpret(FD{T, f}, quotient * powt + round(T, remainder / yi * powt))
 end
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -143,6 +143,10 @@ function /{T, f}(x::FD{T, f}, y::FD{T, f})
     reinterpret(FD{T, f}, quotient * powt + round(T, remainder / y.i * powt))
 end
 
+# these functions are needed to avoid InexactError when converting from the integer type
+/{T, f}(x::Integer, y::FD{T, f}) = reinterpret(FD{T, f}, round(T, x / y.i))
+/{T, f}(x::FD{T, f}, y::Integer) = reinterpret(FD{T, f}, round(T, x.i / y))
+
 # integerification
 trunc{T, f}(x::FD{T, f}) = FD{T, f}(div(x.i, T(10)^f))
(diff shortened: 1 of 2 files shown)


src/FixedPointDecimals.jl
```diff
@@ -143,6 +145,21 @@ function /{T, f}(x::FD{T, f}, y::FD{T, f})
     reinterpret(FD{T, f}, quotient * powt + round(T, remainder / y.i * powt))
 end
 
+# these functions are needed to avoid InexactError when converting from the integer type
+function /{T, f}(x::Integer, y::FD{T, f})
+    powt = T(10)^f
+    xi, yi = checked_mul(x, powt), y.i
+    quotient, remainder = divrem(xi, yi)
+    reinterpret(FD{T, f}, quotient * powt + round(T, remainder / yi * powt))
```
TotalVerb: I'm having trouble seeing why situations where `InexactError` is thrown don't throw `OverflowError` here, since we're doing a checked multiplication with `powt` always.

test/runtests.jl
```diff
@@ -238,6 +244,42 @@ end
     @testset "divide $x by 0" for x in keyvalues[FD2]
         @test_throws DivideError x/FD2(0)
     end
+
+    @testset "rounding" begin
+        # RoundNearest: 1.27 / 2 == 0.635 rounds up to 0.64
+        @test FD2(1.27) / FD2(2) == FD2(0.64)
```
TotalVerb: These tests don't really test the behavior if an overflow would have occurred. In fact I think that they would have passed under the old code also. They're good tests and we should keep them anyway, but we'd also some tests like the `Int8` tests you added above.

src/FixedPointDecimals.jl
omus: I'll remove the`checked_mul` as it isn't necessary yet. I have another PR in the works that deals with overflow when working `powt`.

TotalVerb: Ah, I see; this is keeping the type of the integer passed in. I don't particularly like that; it means this is no better for types that are bigger than or same as machine integer, and that some overflows would be silent. I think a better way to write this would be:

```julia
powt = T(10)^f
xi, yi = widemul(x, powt), widen(y.i)
quotient, remainder = divrem(xi, yi)
reinterpret(FD{T, f}, round(T, quotient * powt + round(T, remainder / widemul(yi, powt)))
```

which should be a lot safer, and won't silently overflow.

test/runtests.jl
omus: These tests using `FD2` are strictly to make sure the RoundNearest behaviour is working. The original code did pass these tests but one revision I had slipped through without these.

src/FixedPointDecimals.jl
TotalVerb: I notice you did switch from `widemul` to `checked_mul`. However, `widemul` should not use `BigInt` unless `Int128`s are coming in, which should be a rare use case, and `widemul` is more general.

omus: The reason I chose not to use `widemul` was specifically because widening to a `BigInt` would result in an intermediate `BigFloat` which would give more accurate results then working with a `Float64`. If we switch to `widemul` we'll need to revise the following test:

```julia
division by 2: Test Failed
  Expression: (typemax(T) / 2 - eps(T)) * 2 == typemax(T) - eps(T)
   Evaluated: FixedDecimal{Int128,2}(1701411834604692317316873037158841057.24) == FixedDecimal{Int128,2}(1701411834604692317316873037158841057.26)
```

TotalVerb: Good point. It would be nice to avoid using floats at all here, but that's for another PR.

omus: What's the verdict for this PR? I'd rather avoid using `widemul` at the moment here to avoid the testcase failure above. I think switch to just using `*` as I'll be dealing with the overflow issues soon enough.

Aside: I'll try to have the overflow PR out tomorrow even if I'm not complete. It's taking some time as test cases can be challenging to write.

TotalVerb: I think it's fine to merge it as long as we don't tag a new release until the overflow problems are fixed. Thanks!

omus:
Thanks for the review. I'll look over this one more time tomorrow and then merge.

PR #7, opened 2017-05-17
Add a basic README

omus:
The initial start to the FixedPointDecimals documentation. Just meant to cover the very basics for the v0.0.1 release.


omus:
> Inconsistent spacing in e.g., FixedDecimal{T <: Integer, f} vs FixedDecimal{Int8,2}

Thanks. Will fix

> I think you should demo parse and discuss differences between constructing from an AbstractFloat vs from a string.

I definitely need to document this but at the moment the [parsing PR](https://github.com/JuliaMath/FixedPointDecimals.jl/pull/1) hasn't been merged. For now I'll leave that out.

> Use something different than the assignment-equals in your first example (maybe ==?)

I thought about that to. I got the idea from the [FixedPointNumbers.jl README](https://github.com/JuliaMath/FixedPointNumbers.jl#type-hierarchy-and-interpretation) which I think uses the single `=` to be more math like. Maybe I'll update this.

PR #8, opened 2017-05-23
Avoid overflow when working with `powt`

omus:
Several of the `FixedDecimal{T,f}` functions used `T(10)^f` to compute `10ᶠ`. This calculation can easily run into overflow especially if the storage type `T` is small and number of places to store `f` is large.

For example, given `T = Int8` which has a `typemax(Int8) = 127` and `f = 3` if we attempt to compute `powt = T(10)^f` we end up with `-24` instead of `1000`. Using the `powt` value during calculations can cause the resulting FixedDecimal to be incorrect which is very bad for library whose purpose is to have more accurate calculations than floating-point.

I've revised the code to make use of a new function `coefficient(::Type{FD{T,f}})` which calculates `powt` without ever overflowing. Additionally, since none of these scenarios were previously tested I added several new testsets to ensure the overflow issues are not reintroduced. Finally, these changes found some additional problems regarding FixedDecimal division which no longer uses an intermediate floating-point number.

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -204,8 +204,9 @@ function round{T, f}(::Type{FD{T, f}}, x::Rational,
 end
 
 # conversions and promotions
-convert{T, f}(::Type{FD{T, f}}, x::Integer) =
-    reinterpret(FD{T, f}, round(T, Base.widemul(T(x), T(10)^f)))
+function convert{T, f}(::Type{FD{T, f}}, x::Integer)
+    reinterpret(FD{T, f}, T(widemul(x, coefficient(FD{T, f}))))
+end
 
 convert{T <: FD}(::Type{T}, x::AbstractFloat) = round(T, x)
 
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -253,8 +253,7 @@ function convert{TI <: Integer, T, f}(::Type{TI}, x::FD{T, f})::TI
     div(x.i, coefficient(FD{T, f}))
 end
 
-convert{TR<:Rational,T,f}(::Type{TR}, x::FD{T, f})::TR =
-    x.i // T(10)^f
+convert{TR <: Rational, T, f}(::Type{TR}, x::FD{T, f})::TR = x.i // coefficient(FD{T, f})
 
 promote_rule{T, f, TI <: Integer}(::Type{FD{T, f}}, ::Type{TI}) = FD{T, f}
 promote_rule{T, f, TF <: AbstractFloat}(::Type{FD{T, f}}, ::Type{TF}) = TF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -72,17 +72,25 @@ for fn in [:trunc, :floor, :ceil]
 end
 
 """
-    FixedDecimal{I <: Integer, f::Int}
+    FixedDecimal{T <: Integer, f::Int}
 
-A fixed-point decimal type backed by integral type `I`, with `f` digits after
+A fixed-point decimal type backed by integral type `T`, with `f` digits after
 the decimal point stored.
 """
 immutable FixedDecimal{T <: Integer, f} <: Real
     i::T
 
-    # internal constructor
-    Base.reinterpret{T, f}(::Type{FixedDecimal{T, f}}, i::Integer) =
-        new{T, f}(i % T)
+    # inner constructor
+    function Base.reinterpret{T, f}(::Type{FixedDecimal{T, f}}, i::Integer)
+        if T != BigInt && 0 <= f <= max_exp10(T) || T == BigInt && f >= 0
+            new{T, f}(i % T)
+        else
+            throw(ArgumentError(
+                "Requested number of decimal places $f exceeds the max allowed for the " *
+                "storage type $T: [0, $(max_exp10(T))]"
+            ))
+        end
+    end
 end
 
 const FD = FixedDecimal
@@ -424,7 +432,13 @@ T or wider.
     end
 end
 
-coefficient{T, f}(::Type{FD{T, f}}) = exp10(FD{T, f})
+"""
+    coefficient(::Type{FD{T, f}}) -> T
+
+Compute `10^f` as an Integer without overflow. Note that overflow will not occur for any
+constructable `FD{T, f}`.
+"""
+coefficient{T, f}(::Type{FD{T, f}}) = T(10)^f
 coefficient{T, f}(fd::FD{T, f}) = coefficient(FD{T, f})
 value(fd::FD) = fd.i
 
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -283,7 +283,7 @@ function print{T, f}(io::IO, x::FD{T, f})
 
     # note: a is negative if x.i == typemin(x.i)
     s, a = sign(x.i), abs(x.i)
-    integer, fractional = divrem(a, T(10)^f)
+    integer, fractional = divrem(a, exp10(T, f))
     integer = abs(integer)  # ...but since f > 0, this is positive
     fractional = abs(fractional)
 
@@ -380,6 +380,17 @@ function parse_round{T}(::Type{T}, fractional::AbstractString, ::RoundingMode{:N
         end
     end
     return T(0)
+end
+
+function exp10_max{T}(::Type{T})
+    length(digits(typemax(T))) - 1
+end
+
+function exp10(T::Type, y::Integer)
+    while T != BigInt && exp10_max(T) < y
+        T = widen(T)
+    end
+    T(10)^y
 end
 
 end
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -408,7 +408,8 @@ T or wider.
     end
 end
 
-coefficient{T, f}(fd::FD{T, f}) = widen(exp10(FD{T, f}))
+coefficient{T, f}(::Type{FD{T, f}}) = exp10(FD{T, f})
+coefficient{T, f}(fd::FD{T, f}) = coefficient(FD{T, f})
 value(fd::FD) = fd.i
 
 end
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -190,11 +190,11 @@ for fn in [:trunc, :floor, :ceil]
         reinterpret(FD{T, f}, val)
     end
 end
-round{TI <: Integer}(::Type{TI}, x::FD,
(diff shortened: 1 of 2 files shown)


src/FixedPointDecimals.jl
```diff
@@ -138,39 +138,40 @@ end
 *{T, f}(x::Integer, y::FD{T, f}) = reinterpret(FD{T, f}, T(x * y.i))
 *{T, f}(x::FD{T, f}, y::Integer) = reinterpret(FD{T, f}, T(x.i * y))
 
-# TODO. this is probably wrong sometimes.
 function /{T, f}(x::FD{T, f}, y::FD{T, f})
-    powt = T(10)^f
+    powt = coefficient(FD{T,f})
     quotient, remainder = divrem(x.i, y.i)
-    reinterpret(FD{T, f}, quotient * powt + round(T, remainder / y.i * powt))
+    reinterpret(FD{T, f}, T(widemul(quotient, powt) + round(T, remainder // y.i * powt)))
```
TotalVerb: The use of `Rational` here is interesting, but I think more expensive than it needs to be. It looks like we really only need to generalize the `_round_to_even` function to accept any number (such as `y.i * powt`) instead of only a power of 10.

src/FixedPointDecimals.jl
```diff
@@ -185,49 +186,48 @@ for fn in [:trunc, :floor, :ceil]
...
 convert{T <: FD}(::Type{T}, x::AbstractFloat) = round(T, x)
 
 function convert{T, f}(::Type{FD{T, f}}, x::Rational)::FD{T, f}
-    powt = T(10)^f
-    num::T, den::T = numerator(x), denominator(x)
-    g = gcd(powt, den)
-    powt = div(powt, g)
-    den = div(den, g)
-    reinterpret(FD{T, f}, powt * num) / FD{T, f}(den)
+    powt = coefficient(FD{T, f})
```
TotalVerb: Could you explain this change? It seems to me that `x * powt` is susceptible to overflow, and the original code was better at avoiding that.

src/FixedPointDecimals.jl
```diff
@@ -311,4 +315,34 @@ function show{T, f}(io::IO, x::FD{T, f})
...
 end
 
+
+"""
+    max_exp10(T)
+
+The highest value of `x` which does not result in an overflow when evaluating `T(10)^x`.
+"""
+function max_exp10{T <: Integer}(::Type{T})
+    length(digits(typemax(T))) - 1
```
TotalVerb: This allocates memory, which we should avoid. There are better way to compute this without overflow.

However I would rather, personally, ensure that 1 is representable in every `FD` type. That means requiring `typemax(IntType) >= 10^f` for every constructible `FD{IntType, f}` type, which would not require excessive overflow checks which I would like to avoid in Currencies. Is there a use case for `FixedDecimal` types that cannot represent `1`? (Note that this change almost makes all operations on those types `InexactError`s anyway, so I lean toward simply disallowing their construction.)

src/FixedPointDecimals.jl
TotalVerb: Well, I suppose since we are using `widemul` anyway, it could make sense to use the wider rational type.

omus: I ended up switching to the `Rational` to fix a precision problem. I'll verify that this is necessary as I may have just been running into this issue at the time: https://github.com/JuliaLang/julia/pull/19779

TotalVerb: Right, this should be precise where the original floating point isn't. However, see my PR #9 which achieves the same result but with no possibility of intermediate overflow.

src/FixedPointDecimals.jl
omus: I changed since some FixedDecimal numbers with large values of `f` may not be able to represent the `den`. The `x * powt` seems to be safe from silently overflowing as Rationals will either raise an `OverflowError` or promote correctly:

```julia
julia> (typemax(Int) // 1) * 2
ERROR: OverflowError()
 in *(::Rational{Int64}, ::Rational{Int64}) at ./rational.jl:208
 in *(::Rational{Int64}, ::Int64) at ./promotion.jl:191

julia> (typemax(Int) // 1) * BigInt(2)
18446744073709551614//1
```

TotalVerb: Can we keep the old code for now, replacing `/ FD{T, f}(den)` with `/ den` which avoids the issue with representation of `den`? It won't silently overflow, but this code looks nevertheless susceptible to `OverflowError` in more cases than before. We should revisit it later.

TotalVerb: Actually, I see the difference now: the old code handled inexact conversions, whereas this one will throw `InexactError`. I'm OK with the change to throw `InexactError` as it does match what happens if you convert a fixed-decimal with higher precision to one with lower precision; we should preserve the old code as a `round` method.

src/FixedPointDecimals.jl
omus: I agree that making sure 1 can always be represented in every `FD` type is reasonable. I personally, don't have a use case for this but I my original thought was that `FixedDecimal`s purpose is for representing decimal numbers without losing precision so it made sense to support a decimal at any arbitrary precision. Your proposal also implies that we should not allow any value of `f < 0`.

As for "excessive overflow checks" we're still going to need to make use of `widemul` in quite a few places as we can still run into overflow with large integers:

```julia
julia> T = Int; powt = T(10)^2
100

julia> typemax(T) * powt
-100
```

TotalVerb: `widemul` should generally (except in the `BigInt` case) be much cheaper than the overflow checks for exponentiation as implemented here, so I'm OK with that.

omus: Ok, I'll switch the `coefficient` function to just be `T(10)^f` and add the appropriate restrictions when constructing a FixedDecimal instance.

omus:
@TotalVerb the main driver behind these changes is making sure that `FixedDecimal` arithmetic doesn't result in silent overflow. Do you agree that the FixedPointDecimals package should be robust against issues like overflow or implicit loss of precision?

TotalVerb:
Here's my stance:

- Addition and subtraction of `FixedDecimal`s is currently no more expensive than of integers, which is a property that `Currencies` does really want. So I would be against overflow checking here.
- On the other hand, I'm fine with using `widemul` for multiplication/division because the cost of the integer division should dwarf the checks for overflow and the use of a bigger integer type, even `Int128`. But `digits` is substantially more expensive than everything, so that would be a no-go for overflow checks. Luckily it doesn't seem necessary.
- Regarding implicit loss of precision, I agree in general for converting to `FixedDecimal`, though operations like multiplication and division do implicitly lose precision (through rounding), and I'm fine with that as long as we round correctly at all times.

TotalVerb:
To provide some more context, here's a quick benchmark:

```julia
julia> X = collect(1:100);

julia> Y = collect(1:100);

julia> @btime $X .+ $Y;
  56.707 ns (1 allocation: 896 bytes)

julia> @btime Base.Checked.checked_add.($X, $Y);
  157.888 ns (1 allocation: 896 bytes)

julia> @btime widemul.($X, $Y);
  163.220 ns (1 allocation: 1.77 KiB)

julia> @btime $X .* $Y;
  68.801 ns (1 allocation: 896 bytes)

julia> @btime div.($X, $Y);
  800.957 ns (1 allocation: 896 bytes)

julia> @btime divrem.($X, $Y);
  1.446 μs (1 allocation: 1.77 KiB)

julia> @btime 10 .^ $X;
  528.085 ns (1 allocation: 896 bytes)

julia> @btime digits.($X);
  5.047 μs (101 allocations: 10.27 KiB)
```

some observations:

- checked_add is almost three times slower than +, so in addition I would prefer to avoid it
- `widemul` is almost three times slower than `*`, BUT in multiplication we're doing a `divrem` which is much worse, so I'm happy to use it (relatively speaking, it's not that expensive)
- `10 ^ x` isn't cheap but not worse than integer division
- `digits` is too expensive for most operations, so we should not use that

omus:
> `digits` is too expensive for most operations, so we should not use that

I'll note that the `digits` function is only being used within `exp10(::Type{FD{T,f}})` which is a generated function and only be called once per distinct `FD{T,f}` type.

TotalVerb:
@omus Ah, yeah, that's fine then. We should replace this generated function with a `Base.@pure` function if possible, as that should also be evaluated at compile time, with less overhead.

omus:
I ended up using `max_exp10` in the FixedDecimal constructor since `typemax(IntType) >= 10^f` has overflow issues. I ended up improving the performance of `max_exp10` which makes the performance comparable to `T(10)^f` with the exception of `Int128` and `UInt128` which widen to `BigInt`s.

src/FixedPointDecimals.jl
```diff
@@ -382,4 +394,36 @@ function parse_round{T}(::Type{T}, fractional::AbstractString, ::RoundingMode{:N
     return T(0)
 end
 
+
+"""
+    max_exp10(T)
+
+The highest value of `x` which does not result in an overflow when evaluating `T(10)^x`.
+"""
+function max_exp10{T <: Integer}(::Type{T})
```
TotalVerb: We can do a branch here so that

```
if !applicable(typemax, T)
    return typemax(Int)
end
```

which would be a little nicer than the current approach in that it will handle arbitrary-precision integers that are not `BigInt` properly (e.g. if someone had their own custom implementation).

omus: I think I would rather do:
```julia
if !applicable(typemax, T)
    throw(DomainError())
end
```
or alternatively
```julia
if !applicable(typemax, T)
    return -1
end
```
Returning a large exponent which is guaranteed to overflow doesn't seem like the right answer.

TotalVerb: But `typemax(Int)` won't "overflow" for `BigInt` or other types that don't define `typemax`. It would just simplify the implementation of the constructor so as not to require special-casing `BigInt`.

omus: What I mean is that `T(10)^x` where T is BigInt and `x = typemax(Int)` does actually overflow:
```julia
julia> BigInt(10)^typemax(Int)
gmp: overflow in mpz type

signal (6): Abort trap: 6
while loading no file, in expression starting on line 0
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 3959986 (Pool: 3958822; Big: 1164); GC: 25
Abort trap: 6
```
I agree with your point that the FixedDecimal constructor shouldn't have the BigInt special case so I'll try to generalize the code in a different way.

omus:
> I think we can remove the BigInt special case. Otherwise, LGTM (after a rebase). Thanks!

I'll push the changes and let you review before I rebase.

omus:
I'm going to do a merge instead of our typical rebase.

PR #9, opened 2017-05-23
Fix overflow issues with division

TotalVerb:
Here is an alternative approach to fixing division that I think is cleaner (it's mostly orthogonal to the changes you propose in #8, but replaces the use of rational rounding, which as far as I can tell is susceptible to overflow errors that could be avoided). We may end up widening `Int64` to `BigInt` for the `Int64 / FD{Int64, f}` case, but that is hard to avoid while still being able to evaluate all divisions.

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -34,8 +34,6 @@ import Base: reinterpret, zero, one, abs, sign, ==, <, <=, +, -, /, *, div,
              typemin, typemax, realmin, realmax, print, show, string, convert,
              promote_rule, min, max, trunc, round, floor, ceil, eps, float, widemul
 
-import Base.Checked: checked_mul
-
 const IEEEFloat = Union{Float16, Float32, Float64}
 
 for fn in [:trunc, :floor, :ceil]
@@ -109,18 +107,22 @@ Base.@pure function widemul{T, f}(x::FD{T, f}, y::Integer)
 end
 Base.@pure widemul(x::Integer, y::FD) = widemul(y, x)
 
-function _round_to_even(quotient, remainder, powt)
-    if powt == 1
-        quotient
+"""
+    _round_to_even(quotient, remainder, divisor)
+
+Round `quotient + remainder / divisor` to the nearest even integer, given that
+`0 ≤ remainder < divisor` or `0 ≥ remainder > divisor`. (This assumption is
+satisfied by the return value of `fldmod` in all cases, and the return value of
+`divrem` in cases where `divisor` is known to be positive.)
+"""
+function _round_to_even(quotient, remainder, divisor)
+    halfdivisor = divisor >> 1
+    if iseven(divisor) && remainder == halfdivisor
+        ifelse(iseven(quotient), quotient, quotient + one(quotient))
+    elseif abs(remainder) > abs(halfdivisor)
+        quotient + one(quotient)
     else
-        halfpowt = powt >> 1
-        if remainder == halfpowt
-            ifelse(iseven(quotient), quotient, quotient + one(quotient))
-        elseif remainder < halfpowt
-            quotient
-        else
-            quotient + one(quotient)
-        end
+        quotient
     end
 end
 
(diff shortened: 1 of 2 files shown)


src/FixedPointDecimals.jl
```diff
@@ -138,26 +140,33 @@ end
...
+    xi, yi = promote(x, y.i)
+
+    # The integer part of our result is x.i * 10^2f / y.i, so we need to
+    # double-widen to get a precise result.
+    powt = S(10)^f
+    powtsq = widemul(powt, powt)
+    quotient, remainder = fldmod(widemul(widen(xi), powtsq), widen(widen(yi)))
+
+    reinterpret(FD{T, f},
+                T(_round_to_even(quotient, remainder, widen(widen(yi)))))
```
omus: What number of characters do you word wrap to? This line can fit entirely in a [92-character line limit](https://docs.julialang.org/en/stable/manual/documentation/#documentation) which is the Julia standard

src/FixedPointDecimals.jl
```diff
@@ -138,26 +140,33 @@ end
...
-    quotient, remainder = divrem(xi, yi)
-    reinterpret(FD{T, f}, quotient * powt + round(T, remainder / yi * powt))
+    S = promote_type(typeof(x), T)
+    xi, yi = promote(x, y.i)
+
+    # The integer part of our result is x.i * 10^2f / y.i, so we need to
+    # double-widen to get a precise result.
+    powt = S(10)^f
+    powtsq = widemul(powt, powt)
+    quotient, remainder = fldmod(widemul(widen(xi), powtsq), widen(widen(yi)))
```
omus: The `widen` calls here seem unnecessary. Isn't this code the same as:
```julia
powt = T(10)^f
powtsq = widemul(powt, powt)
quotient, remainder = fldmod(widemul(x, powtsq), y.i)
reinterpret(FD{T, f}, T(_round_to_even(quotient, remainder, y.i)))
```
Given that `T(10)^f` doesn't silently overflow.

src/FixedPointDecimals.jl
```diff
@@ -138,26 +140,33 @@ end
...
+    reinterpret(FD{T, f},
+                T(_round_to_even(quotient, remainder, widen(widen(yi)))))
 end
 
 function /{T, f}(x::FD{T, f}, y::Integer)
-    powt = T(10)^f
-    xi, yi = x.i, checked_mul(y, powt)
-    quotient, remainder = divrem(xi, yi)
-    reinterpret(FD{T, f}, quotient * powt + round(T, remainder / yi * powt))
+    S = promote_type(T, typeof(y))
```
omus: Delete

omus:
@TotalVerb did some refactoring. Would be good if you could verify the changes.

src/FixedPointDecimals.jl
```diff
@@ -142,31 +142,21 @@ end
...
 end
 
 function /{T, f}(x::FD{T, f}, y::Integer)
-    S = promote_type(T, typeof(y))
-    xi, yi = promote(x.i, y)
-
-    quotient, remainder = fldmod(xi, yi)
-    reinterpret(FD{T, f}, T(_round_to_even(quotient, remainder, yi)))
+    quotient, remainder = fldmod(x.i, y)
+    reinterpret(FD{T, f}, T(_round_to_even(quotient, remainder, y)))
```
TotalVerb: I'm a little worried that passing heterogenous integer types to `_round_to_even` is a bad idea. It looks like it will work, though?

Personally I'm uncomfortable with relying on automatic promotion, but since this does seem to work and simplifies code, it lgtm.

omus: Good point. What do you think if we would have a `promote` call in `_round_to_even`?

TotalVerb: That would certainly make me more comfortable with it 👍.

TotalVerb: Actually, I think it might be easier just to make `_round_to_even` dispatch only on homogenous types, and have a promoting fallback.

TotalVerb: I pushed a quick (untested) change from the web interface; hopefully this will pass CI.

PR #13, opened 2017-07-25
remove base method extension in tests

tkelman:
unless the tests aren't covering something (or it's platform-dependent),
this doesn't seem to be necessary for them to pass?

ref https://github.com/JuliaMath/FixedPointDecimals.jl/commit/fa45e3631116c31b06e4d8e510d0520a0460c2b7#commitcomment-23278547

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -72,13 +72,6 @@ function parse_int{T, f}(::Type{FD{T, f}}, val::AbstractString; ceil::Bool=false
     reinterpret(FD{T, f}, parse(T, val[1:(f + 1)]) + T(ceil))
 end
 
-# When working with typemax of Int128 or UInt128 we can run into issues.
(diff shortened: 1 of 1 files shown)


omus:
> unless the tests aren't covering something (or it's platform-dependent),
this doesn't seem to be necessary for them to pass?

I believe the tests used to require this but after several rounds of refactoring I believe we managed to remove this requirement. Thanks for catching this.

tkelman:
My initial local tests of this I made a mistake and wasn't using 0.5 where it would have made a difference. But going back, at the commit these lines were added it was needed on 0.5, on master it's no longer needed.

omus:
I tested against Julia 0.5.0 - 0.5.2 and couldn't generate a test failure. On which version did you discover this fix was needed?

tkelman:
checking out fa45e3631116c31b06e4d8e510d0520a0460c2b7 on 0.5.2

PR #14, opened 2017-10-24
Use only bits types for CONTAINER_TYPES

omus:
A change introduced in Julia 0.7 made BigInt a subtype of Signed (JuliaLang/julia#23473). The change resulted in the code including BigInt in the CONTAINER_TYPES which resulted in an exceptions since BigInt is not a bits type.

Note that the JuliaLang/julia#23473 also introduced [a bug](https://github.com/JuliaLang/julia/issues/24298) which is causing the tests to fail.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -17,7 +17,7 @@ const FD4 = FixedDecimal{Int, 4}
 const WFD2 = FixedDecimal{Int128, 2}
 const WFD4 = FixedDecimal{Int128, 4}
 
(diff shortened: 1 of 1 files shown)


TotalVerb:
Which tests do the BigInt container type fail?

omus:
Tests that fail are currently "multiply: limits of UInt128" and "division: limits of UInt128" only on Julia 0.7. I've already made a fix JuliaLang/julia#24313

PR #15, opened 2017-10-24
Fix most Julia 0.7 deprecations

omus:
Fixed only the deprecations that retain compatibility with Julia 0.5. I'll switch to Julia 0.6 in another PR.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,7 +1,7 @@
 using FixedPointDecimals
 import FixedPointDecimals: FD, value
-using Base.Test
 using Compat
+using Compat.Test
 import Base.Checked: checked_mul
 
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -1,12 +1,15 @@
 using FixedPointDecimals
 import FixedPointDecimals: FD, value
-using Base.Test
 using Compat
 import Base.Checked: checked_mul
 
-include("utils.jl")
+if VERSION >= v"0.7.0-DEV.2047"
```
TotalVerb: the recommendation is `using Compat.Test` for now.

omus: Thanks. I missed that one.

PR #18, opened 2017-12-06
Avoid redundant call to max_exp10(T)

omus:
Trivial change which makes use of the already computed `max_exp10(T)` stored in `n`.


omus:
Will merge when the CI finishes (if ever). After this is merged I'll also tag a new release.

PR #19, opened 2017-12-22
Fix `var` and `std` when using 0 or 1-element array

omus:
Makes the `var` and `std` calculations always work.

Before PR
```julia
julia> using FixedPointDecimals

julia> std([FixedDecimal{Int,1}(384.0)])
ERROR: DivideError: integer division error
```

After PR
```julia
julia> using FixedPointDecimals

julia> std([FixedDecimal{Int,1}(384.0)])
NaN
```

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -30,7 +30,7 @@ export FixedDecimal, RoundThrows
 using Compat
 
 import Base: reinterpret, zero, one, abs, sign, ==, <, <=, +, -, /, *, div, rem, divrem,
-             fld, mod, fldmod, fld1, mod1, fldmod1, isinteger, typemin, typemax,
+             fld, mod, fldmod, fld1, mod1, fldmod1, var, isinteger, typemin, typemax,
              realmin, realmax, print, show, string, convert, parse, promote_rule, min, max,
              trunc, round, floor, ceil, eps, float, widemul
 
@@ -176,6 +176,15 @@ end
 function /(x::FD{T, f}, y::Integer) where {T, f}
     quotient, remainder = fldmod(x.i, y)
     reinterpret(FD{T, f}, T(_round_to_even(quotient, remainder, y)))
+end
+
+# Use floating-point when calculating variance to avoid divide by zero errors
+function var(A::AbstractArray{<:FixedDecimal}; corrected::Bool=true, mean=nothing)
+    var(convert.(AbstractFloat, A); corrected=corrected, mean=mean)
+end
+
+function var(A::AbstractArray{<:FixedDecimal}, region; corrected::Bool=true, mean=nothing)
+    var(convert.(AbstractFloat, A), region; corrected=corrected, mean=mean)
 end
 
 # integerification
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -494,6 +494,16 @@ end
     end
 end
 
+@testset "var" begin
+    @test isequal(var(FD2[]), NaN)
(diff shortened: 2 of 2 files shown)


TotalVerb:
Isn't it a little strange to return a Float when the input type isn't?

omus:
I'm not super happy about returning a Float either. An alternative to this would be to introduce a concept of `NaN` (and probably also `Inf` and `-Inf`) into `FixedDecimal`.

TotalVerb:
I'm not sure that that's a sensible change, since fixed point numbers behave like integers in most cases. By the way, it is also the case that

```julia
julia> std([1//1])
ERROR: DivideError: integer division error
Stacktrace:
 [1] //(::Rational{Int64}, ::Rational{Int64}) at ./rational.jl:51
 [2] #varm#533(::Bool, ::Function, ::Array{Rational{Int64},1}, ::Rational{Int64}) at ./statistics.jl:153
 [3] (::Base.#kw##varm)(::Array{Any,1}, ::Base.#varm, ::Array{Rational{Int64},1}, ::Rational{Int64}) at ./<missing>:0
 [4] #var#536(::Bool, ::Void, ::Function, ::Array{Rational{Int64},1}) at ./statistics.jl:184
 [5] (::Base.#kw##var)(::Array{Any,1}, ::Base.#var, ::Array{Rational{Int64},1}) at ./<missing>:0
 [6] std(::Array{Rational{Int64},1}) at ./statistics.jl:246

julia> var([1//1])
ERROR: DivideError: integer division error
Stacktrace:
 [1] //(::Rational{Int64}, ::Rational{Int64}) at ./rational.jl:51
 [2] #varm#533(::Bool, ::Function, ::Array{Rational{Int64},1}, ::Rational{Int64}) at ./statistics.jl:153
 [3] (::Base.#kw##varm)(::Array{Any,1}, ::Base.#varm, ::Array{Rational{Int64},1}, ::Rational{Int64}) at ./<missing>:0
 [4] #var#536(::Bool, ::Void, ::Function, ::Array{Rational{Int64},1}) at ./statistics.jl:184
 [5] var(::Array{Rational{Int64},1}) at ./statistics.jl:184
```

omus:
Counter example:
```julia
julia> std(Int[])
NaN

julia> std(Int[1])
NaN
```

TotalVerb:
OK, that's fair. I'm fine with returning a float if other cases do so also.

omus:
Although `FixedDecimal`s mostly work the same way as Integers they still represent a subset of real values and can be treated as such. As I said I'm not crazy about switching to Floats so I think an alternative proposal would be to update this PR to throw a better error message in the length < 2 cases where we attempt to return a value which can't be represented as a `FixedDecimal`.

omus:
Which ever way we end up going on this I'm ok with waiting on this PR for a while to choose a good direction.

TotalVerb:
I just realized that Rational does behave this way already, except in the one-element case:

```julia
julia> std([1//1, 1//2])
0.3535533905932738

julia> typeof(ans)
Float64
```

In this case I would expect that the Base rational type be fixed to just allow the one-element case, since the result is a float anyway.

omus:
The same is true for `FixedDecimal` before this PR:
```julia
julia> std(FixedDecimal{Int,2}[1, 2])
0.7071067811865476

julia> typeof(ans)
Float64
```
With this in mind I think I may be able to improve this fix.

omus:
Should be partially addressed by https://github.com/JuliaLang/julia/pull/25304. We may also be able to add some logic into Base to return a `NaN` instead of throwing an error in `var` which would replace the changes here.

PR #20, opened 2017-12-28
Updates for Julia 0.7.0-DEV.3199

omus:


src/FixedPointDecimals.jl
```diff
@@ -100,6 +100,10 @@ struct FixedDecimal{T <: Integer, f} <: Real
     end
 end
 
+function (::Type{FixedDecimal{T, f}})(x::Real) where {T, f}
```
TotalVerb: I would write this as `(::Type{T})(x::Real) where {T <: FixedDecimal} = convert(T, x)`.

PR #21, opened 2018-02-21
Add hashing support

rofinn:
@TotalVerb or @ararslan: Would you be willing to review this change? I don't have permission to merge this and Curtis isn't available.

NOTE: This package already wasn't passing tests on nightly.

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -473,8 +473,6 @@ coefficient(::Type{FD{T, f}}) where {T, f} = T(10)^f
 coefficient(fd::FD{T, f}) where {T, f} = coefficient(FD{T, f})
 value(fd::FD) = fd.i
 
-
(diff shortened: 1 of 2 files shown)


rofinn:
Thank you.

TotalVerb:
This should use the `Rational` hash instead of `BigFloat` one? That would be less susceptible to BigFloat precision loss.

```
julia> FD{Int,2}(1//10) == 1//10
true

julia> hash(FD{Int,2}(1//10)) == hash(1//10)
┌ Warning: Constructors no longer fall back to `convert`. A constructor `BigFloat(::FixedDecimal{Int64,2})` should be defined instead.
│   caller = hash(::FixedDecimal{Int64,2}, ::UInt64) at FixedPointDecimals.jl:478
└ @ FixedPointDecimals FixedPointDecimals.jl:478
false
```

Sorry I didn't catch this earlier. I'll PR a fix.

PR #22, opened 2018-02-25
Fix hash support

TotalVerb:
Address some small issues in the hash function from #21.

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -32,7 +32,7 @@ using Compat
 import Base: reinterpret, zero, one, abs, sign, ==, <, <=, +, -, /, *, div, rem, divrem,
              fld, mod, fldmod, fld1, mod1, fldmod1, isinteger, typemin, typemax,
              realmin, realmax, print, show, string, convert, parse, promote_rule, min, max,
-             trunc, round, floor, ceil, eps, float, widemul
+             trunc, round, floor, ceil, eps, float, widemul, decompose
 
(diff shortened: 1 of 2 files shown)


src/FixedPointDecimals.jl
```diff
@@ -475,6 +475,7 @@ coefficient(::Type{FD{T, f}}) where {T, f} = T(10)^f
 coefficient(fd::FD{T, f}) where {T, f} = coefficient(FD{T, f})
 value(fd::FD) = fd.i
 
-Base.hash(fd::FD{T, f}, h::UInt) where {T, f} = Base.hash(BigFloat(fd), h)
+# for generic hashing
+decompose(fd::FD) = decompose(Rational(fd))
```
rofinn: Actually, do we want to differentiate between `FixedPointDecimal`s and other numeric types (e.g., `hash(Rational(fd), hash(:FixedPointDecimal, h)`)?

test/runtests.jl
```diff
@@ -947,6 +947,8 @@ end
         @test hash(fd1) == hash(fd2)
         @test hash(fd1) != hash(fd3)
         @test hash(fd1) != hash(fd1.i)
+        @test hash(FD2(1//10)) == hash(1//10)
```
rofinn: Wondering if this should be `hash(FD2(1//10)) != hash(1//10)`? See comment above.

TotalVerb:
No, numbers that are equal in Julia should hash to the same value. [More specifically, isequal and hash must be consistent.]

PR #25, opened 2018-07-07
Need to use 'something()' rather than 'coalesce()' because findfirst(…

macd:
…) returns 'nothing' when there is no match'

It is still upgrade Friday here and so this is a small fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
   - nightly
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
   - linux
   - osx
 julia:
+  - 0.7
   - nightly
 notifications:
   email: false
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,5 @@
 environment:
   matrix:
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe"
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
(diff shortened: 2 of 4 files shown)


macd:
OK, it looks like I forgot to update the CI specs.  But I am not at all certain about how to do that.

TotalVerb:
lgtm

omus:
Besides the `ndigits` change it would be simple to be compatible with Julia 0.6 via Compat.jl

macd:
So the recommendation is to drop 0.6 support.  See https://discourse.julialang.org/t/ann-introducing-upgradathon-fridays/12047/3

Eventually, we will need Project.toml files too, but it is recommended to hold off on those for now.  I will update the REQUIRE to julia 0.7-beta in the PR

macd:
I have looked at the above failures and it is a mystery to me.  Any thoughts from those who know more?

omus:
Compat is still used by the tests.

omus:
I understand that for packages which haven't transitioned at all that dropping 0.6 is easiest. In this case I would say dropping 0.6 is more work

.travis.yml
```diff
@@ -4,7 +4,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
   - nightly
```
omus: Should also add `0.7`

appveyor.yml
```diff
@@ -1,7 +1,5 @@
 environment:
   matrix:
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe"
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x64/julia-latest-win64.exe"
```
omus: Should also add Julia 0.7 here. If your so inclined we can also update the file to look like: https://github.com/JuliaCI/Appveyor.jl/blob/master/appveyor.yml

macd:
Great!  Thanks.  Glad you're doing the 0.6 work since it's been over 6 months since I've had that on my machine.

PR #26, opened 2018-07-26
Support Julia 0.6 again

omus:
Since it's pretty easy to keep supporting at this time why not keep supporting 0.6?


omus:
https://github.com/JuliaLang/Compat.jl/issues/600 would have been useful here

PR #30, opened 2018-10-26
Improve constructor performance

omus:
Performance tested on Julia 1.0. 

Before:
```julia
julia> for i in 1:8
          @btime FixedDecimal{Int,$i}.(1:100)
       end
  8.986 μs (3 allocations: 960 bytes)
  9.034 μs (3 allocations: 960 bytes)
  10.391 μs (3 allocations: 960 bytes)
  9.667 μs (3 allocations: 960 bytes)
  10.601 μs (3 allocations: 960 bytes)
  10.630 μs (3 allocations: 960 bytes)
  10.768 μs (3 allocations: 960 bytes)
  9.423 μs (3 allocations: 960 bytes)
```
After (without pre-computing `max_exp10`):
```julia
julia> for i in 1:8
          @btime FixedDecimal{Int,$i}.(1:100)
       end
  531.746 ns (3 allocations: 960 bytes)
  569.186 ns (3 allocations: 960 bytes)
  1.888 μs (3 allocations: 960 bytes)
  706.234 ns (3 allocations: 960 bytes)
  1.934 μs (3 allocations: 960 bytes)
  2.056 μs (3 allocations: 960 bytes)
  2.153 μs (3 allocations: 960 bytes)
  700.782 ns (3 allocations: 960 bytes)
```
After (with pre-computing `max_exp10`):
```julia
julia> for i in 1:8
          @btime FixedDecimal{Int,$i}.(1:100)
       end
  390.299 ns (3 allocations: 960 bytes)
  426.136 ns (3 allocations: 960 bytes)
  1.875 μs (3 allocations: 960 bytes)
  541.351 ns (3 allocations: 960 bytes)
  1.961 μs (3 allocations: 960 bytes)
  1.966 μs (3 allocations: 960 bytes)
  2.189 μs (3 allocations: 960 bytes)
  533.450 ns (3 allocations: 960 bytes)
```
One caveat is that custom `Integer` types that don't support `typemax` will either need to implement `max_exp10(::Type{MyInt}) = -1` or they will fail when calling the fallback `max_exp10` method.

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -460,7 +460,6 @@ The highest value of `x` which does not result in an overflow when evaluating `T
 types of `T` that do not overflow -1 will be returned.
 """
 function max_exp10(::Type{T}) where {T <: Integer}
-    applicable(typemax, T) || return -1
     W = widen(T)
     type_max = W(typemax(T))
 
@@ -474,6 +473,13 @@ function max_exp10(::Type{T}) where {T <: Integer}
     end
 
     exponent - 1
+end
+
+max_exp10(::Type{BigInt}) = -1
+
+# Pre-compute max_exp10 for the standard Integer types
+for T in Base.BitInteger_types
+    @eval max_exp10(::Type{$T}) = $(max_exp10(T))
 end
 
 """
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -94,12 +94,18 @@ struct FixedDecimal{T <: Integer, f} <: Real
         if f >= 0 && (n < 0 || f <= n)
             new{T, f}(i % T)
         else
-            throw(ArgumentError(
-                "Requested number of decimal places $f exceeds the max allowed for the " *
-                "storage type $T: [0, $n]"
-            ))
+            # Note: introducing a function barrier to improve performance
+            # https://github.com/JuliaMath/FixedPointDecimals.jl/pull/30
+            _throw_storage_error(f, T, n)
         end
     end
+end
+
+@noinline function _throw_storage_error(f, T, n)
+    throw(ArgumentError(
+        "Requested number of decimal places $f exceeds the max allowed for the " *
+        "storage type $T: [0, $n]"
+    ))
 end
 
 const FD = FixedDecimal
@@ -476,11 +482,6 @@ function max_exp10(::Type{T}) where {T <: Integer}
 end
 
 max_exp10(::Type{BigInt}) = -1
-
-# Pre-compute max_exp10 for the standard Integer types
-for T in Base.BitInteger_types
-    @eval max_exp10(::Type{$T}) = $(max_exp10(T))
-end
 
 """
     coefficient(::Type{FD{T, f}}) -> T
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -109,6 +109,20 @@ end
         x = FixedPointDecimals.max_exp10(T)
         @test T(10)^x == widen(T(10))^x
     end
+
+    @testset "custom integer types" begin
+        @eval begin
+            primitive type Int24 <: Integer 24 end
+            Base.typemax(::Type{Int24}) = 2^24
+            Base.widen(::Type{Int24}) = Int32
+        end
+
+        @test FixedPointDecimals.max_exp10(Int24) == 7
+
+        # Note: we're just pretending that this is unbounded
+        @eval primitive type IntUnbounded <: Integer 256 end
+        @test_throws MethodError FixedPointDecimals.max_exp10(IntUnbounded)
+    end
 end
 
 # ensure that the coefficient multiplied by the highest and lowest representable values of


omus:
Thanks to @NHDaly who pointed this out on Slack and @KristofferC giving some good suggestions.

omus:
Related: https://github.com/RelationalAI-oss/FixedPointDecimals.jl/pull/1

src/FixedPointDecimals.jl
```diff
@@ -460,7 +460,6 @@ The highest value of `x` which does not result in an overflow when evaluating `T
 types of `T` that do not overflow -1 will be returned.
 """
 function max_exp10(::Type{T}) where {T <: Integer}
-    applicable(typemax, T) || return -1
```
TotalVerb: Could you explain this change a little more? is it intended that `BigInt` be the only supported unbounded integer type?

omus: At the moment in `Base` the only unbounded `Integer` type is `BigInt`. As it turns out the `applicable` call to check for a function declaration is rather expensive and is stopping the compiler from performing constant folding.

This change does require new `Integer` subtypes to declare `FixedPointDecimals.max_exp10(::Type{X}) = -1` in order to be usable with the `FixedDecimal` type. An alternative and better way to address this would be to have `typemax(::Type{BigInt}) = nothing` defined in Julia itself.

omus:
The coverage reduction is just this line being missed:
https://github.com/JuliaMath/FixedPointDecimals.jl/blob/98e65b6eb890d666255fdf57ada0d905bae2ec23/src/FixedPointDecimals.jl#L478
That line is definitely being tested by:
https://github.com/JuliaMath/FixedPointDecimals.jl/blob/98e65b6eb890d666255fdf57ada0d905bae2ec23/test/runtests.jl#L106
So this looks to just be an issue with coverage reporting.

TotalVerb:
@omus I see. If the compiler is able to do constant folding, does precomputation provide additional benefit?

NHDaly:
> Im surprised precomputing has an impact considering

I was assuming the first numbers simply included compiling in their timings?

NHDaly:
So, I opened https://github.com/JuliaMath/FixedPointDecimals.jl/issues/35 to discuss this more. I think traits would probably be a better solution than this `@eval` solution.

BUT in this case, basically I actually think we can avoid all this and just remove the `applicable` call. It doesn't really add anything...

NHDaly:
So i'm currently proposing this solution: https://github.com/JuliaMath/FixedPointDecimals.jl/pull/34

Of course we need to add tests and stuff like you've done here curtis. Or, if you want, we can keep this PR as the main one, and I can propose changes into it, or w/e you're comfortable with! :) Thanks for taking an interest after reading the Slack thread!! ❤️ !

omus:
> Im surprised precomputing has an impact

I was also surprised. It seems pretty consistent though.

> I was assuming the first numbers simply included compiling in their timings?

`@btime` runs the functions multiple times and shouldn't be effecting the results that much. I ran all of the above benchmarks with a fresh Julia 1.0 session.

NHDaly:
> @btime runs the functions multiple times and shouldn't be effecting the results that much. I ran all of the above benchmarks with a fresh Julia 1.0 session.

Oh yeah, duh. Sorry i missed that. It is weird then..

KristofferC:
The following diff

```diff
diff --git a/src/FixedPointDecimals.jl b/src/FixedPointDecimals.jl
index 385f5df..8debb38 100644
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -94,14 +94,17 @@ struct FixedDecimal{T <: Integer, f} <: Real
         if f >= 0 && (n < 0 || f <= n)
             new{T, f}(i % T)
         else
-            throw(ArgumentError(
-                "Requested number of decimal places $f exceeds the max allowed for the " *
-                "storage type $T: [0, $n]"
-            ))
+            __throw_argerror(f, T, n)
         end
     end
 end

+
+@noinline __throw_argerror(f, T, n) = throw(ArgumentError(
+    "Requested number of decimal places $f exceeds the max allowed for the " *
+    "storage type $T: [0, $n]"
+))
+
```

gives a pretty good speedup.

KristofferC:
I also don't get any speedup from the `@eval` on master.

NHDaly:
> I also don't get any speedup from the `@eval` on master.

oh weird, really? Are you sure? What do you mean "master" -- master julia?

I definitely do get the speedups from this PR:
```
# This is all after checking out the `cv/constructor-performance` branch.
julia> @code_native reinterpret(FixedPointDecimals.FixedDecimal{UInt8,0}, 1)
    .section    __TEXT,__text,regular,pure_instructions
; Function reinterpret {
; Location: FixedPointDecimals.jl:93
    movl    %esi, %eax
    retl
    nopw    %cs:(%eax,%eax)
;}

julia> _test(10_000_000)
  0.082648 seconds (2 allocations: 76.294 MiB, 86.26% gc time)
  0.005308 seconds
  0.023281 seconds (2 allocations: 76.294 MiB, 47.62% gc time)
  0.004669 seconds
```

KristofferC:
I meant on master julia I don't see a difference in performance if the

```
for T in Base.BitInteger_types
    @eval max_exp10(::Type{$T}) = $(max_exp10(T))
end
```

bit is taken out.

NHDaly:
Ah, that's because this PR removes the `applicable` check:
https://github.com/JuliaMath/FixedPointDecimals.jl/pull/30/files#diff-ea4e0198fdaaab3db0fa802fdb33c161L463

And hard-codes the value for `BigInt`:
https://github.com/JuliaMath/FixedPointDecimals.jl/pull/30/files#diff-ea4e0198fdaaab3db0fa802fdb33c161R478

So for a new custom type that doesn't define a `typemax`, it will `MethodError`, whereas before this PR, a custom type would get the `-1` definition.

NHDaly:
> whereas before this PR, a custom type would get the -1 definition.

So maybe that custom definition for `BigInt` could just be changed to a default definition for all types?

omus:
Odd I'm not seeing any speed up with the `__throw_argerror` change. I'm on Julia 1.0.0. I'll try updating.

omus:
> gives a pretty good speedup.

@KristofferC can you share your benchmark and results?

KristofferC:
Heh, so apparently there is only a performance improvement if the 

```
for T in Base.BitInteger_types
    @eval max_exp10(::Type{$T}) = $(max_exp10(T))
end
```

part is removed. Otherwise I guess that the throw part is optimized away.  In other words, the following has the same performance as this PR

```diff
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -94,14 +94,15 @@ struct FixedDecimal{T <: Integer, f} <: Real
         if f >= 0 && (n < 0 || f <= n)
             new{T, f}(i % T)
         else
-            throw(ArgumentError(
-                "Requested number of decimal places $f exceeds the max allowed for the " *
-                "storage type $T: [0, $n]"
-            ))
+            __throw_arg(n, f, T)
         end
     end
 end

+@noinline __throw_arg(n ,f, T) = throw(ArgumentError(
+                "Requested number of decimal places $f exceeds the max allowed for the " *
+                "storage type $T: [0, $n]"
+            ))
 const FD = FixedDecimal

 (::Type{T})(x::Real) where {T <: FD} = convert(T, x)
@@ -477,10 +478,6 @@ end

 max_exp10(::Type{BigInt}) = -1

-# Pre-compute max_exp10 for the standard Integer types
-for T in Base.BitInteger_types
-    @eval max_exp10(::Type{$T}) = $(max_exp10(T))
-end
```

NHDaly:
> In other words, the following has the same performance as this PR

Hmm, no isn't sufficient for me..

```
# with Kristoffer's diff applied:
julia> function _test(max)
         FD_t = FixedPointDecimals.FixedDecimal{Int64, 0}
         sum(1:10)
         sum(FD_t(1):FD_t(10))
         @time X = collect(1:max)
         @time sum(X)
         @time Y = collect(FD_t(1) : FD_t(max))
         @time sum(Y)
         nothing
       end
>> _test (generic function with 1 method)

julia> _test(1) # warmup
  0.000001 seconds (1 allocation: 96 bytes)
  0.000000 seconds
  0.000006 seconds (1 allocation: 96 bytes)
  0.000000 seconds

julia> function _test(max)
         FD_t = FixedPointDecimals.FixedDecimal{Int64, 0}
         sum(1:10)
         sum(FD_t(1):FD_t(10))
         @time X = collect(1:max)
         @time sum(X)
         @time Y = collect(FD_t(1) : FD_t(max))
         @time sum(Y)
         nothing
       end
>> _test (generic function with 1 method)

julia> _test(1) # warmup
  0.000001 seconds (1 allocation: 96 bytes)
  0.000000 seconds
  0.000004 seconds (1 allocation: 96 bytes)
  0.000000 seconds

julia> _test(10_000_000)
  0.054272 seconds (2 allocations: 76.294 MiB, 48.64% gc time)
  0.004916 seconds
  4.221142 seconds (2 allocations: 76.294 MiB, 1.54% gc time)

# It's still doing the runtime checks for `max_exp10`:
julia> @code_lowered reinterpret(FixedPointDecimals.FixedDecimal{UInt8,0}, 1)
>> CodeInfo(
93 1 ─       n = (Main.FixedPointDecimals.max_exp10)($(Expr(:static_parameter, 1)))                                │
94 │   %2  = $(Expr(:static_parameter, 2)) >= 0                                                                    │
   └──       goto #6 if not %2                                                                                     │
   2 ─ %4  = n < 0                                                                                                 │
   └──       goto #4 if not %4                                                                                     │
   3 ─       #temp#@_6 = %4                                                                                        │
   └──       goto #5                                                                                               │
   4 ─       #temp#@_6 = $(Expr(:static_parameter, 2)) <= n                                                        │
   5 ┄       #temp#@_5 = #temp#@_6                                                                                 │
   └──       goto #7                                                                                               │
   6 ─       #temp#@_5 = false                                                                                     │
   7 ┄       goto #9 if not #temp#@_5                                                                              │
95 8 ─ %13 = (Core.apply_type)(Main.FixedPointDecimals.FixedDecimal, $(Expr(:static_parameter, 1)), $(Expr(:static_parameter, 2)))
   │   %14 = $(Expr(:static_parameter, 1))                                                                         │
   │   %15 = i % $(Expr(:static_parameter, 1))                                                                     │
   │   %16 = (Base.convert)(%14, %15)                                                                              │
   │   %17 = %new(%13, %16)                                                                                        │
   └──       return %17                                                                                            │
97 9 ─ %19 = (Main.FixedPointDecimals.__throw_arg)(n, $(Expr(:static_parameter, 2)), $(Expr(:static_parameter, 1)))│
   └──       return %19                                                                                            │
)

```

KristofferC:
Seems julia master is handling this better.

KristofferC:
Also, `@code_lowered` doesn't really tell you anything, use `@code_typed`.

KristofferC:
FWIW, there is a difference between master and julia 1.0.1

master

```
julia> for i in 1:8
                 @btime FixedDecimal{Int,$i}.(1:100)
              end
  464.528 ns (3 allocations: 960 bytes)
  526.536 ns (3 allocations: 960 bytes)
  1.063 μs (3 allocations: 960 bytes)
  1.151 μs (3 allocations: 960 bytes)
  2.135 μs (3 allocations: 960 bytes)
  2.672 μs (3 allocations: 960 bytes)
  1.163 μs (3 allocations: 960 bytes)
  1.064 μs (3 allocations: 960 bytes)
```

1.0.1

```
julia> for i in 1:8
                 @btime FixedDecimal{Int,$i}.(1:100)
              end
  641.452 ns (3 allocations: 960 bytes)
  689.470 ns (3 allocations: 960 bytes)
  2.704 μs (3 allocations: 960 bytes)
  1.070 μs (3 allocations: 960 bytes)
  2.742 μs (3 allocations: 960 bytes)
  2.607 μs (3 allocations: 960 bytes)
  2.864 μs (3 allocations: 960 bytes)
  1.072 μs (3 allocations: 960 bytes)
```

omus:
Function barrier for throw:

Julia 0.6.4

```julia
julia> for i in 1:8
           @btime FixedDecimal{Int,$i}.(1:100)
       end
  51.823 μs (35 allocations: 2.31 KiB)
  52.000 μs (35 allocations: 2.31 KiB)
  52.773 μs (35 allocations: 2.31 KiB)
  52.350 μs (35 allocations: 2.31 KiB)
  52.864 μs (35 allocations: 2.31 KiB)
  53.120 μs (35 allocations: 2.31 KiB)
  53.781 μs (35 allocations: 2.31 KiB)
  53.587 μs (35 allocations: 2.31 KiB)
```

Julia 1.0.1
```julia
julia> for i in 1:8
           @btime FixedDecimal{Int,$i}.(1:100)
       end
  470.204 ns (3 allocations: 960 bytes)
  506.005 ns (3 allocations: 960 bytes)
  1.824 μs (3 allocations: 960 bytes)
  794.865 ns (3 allocations: 960 bytes)
  2.041 μs (3 allocations: 960 bytes)
  1.899 μs (3 allocations: 960 bytes)
  2.104 μs (3 allocations: 960 bytes)
  788.141 ns (3 allocations: 960 bytes)
```

Julia master
```julia
julia> for i in 1:8
          @btime FixedDecimal{Int,$i}.(1:100)
       end
  386.192 ns (3 allocations: 960 bytes)
  419.000 ns (3 allocations: 960 bytes)
  756.138 ns (3 allocations: 960 bytes)
  822.284 ns (3 allocations: 960 bytes)
  1.506 μs (3 allocations: 960 bytes)
  1.929 μs (3 allocations: 960 bytes)
  819.720 ns (3 allocations: 960 bytes)
  815.914 ns (3 allocations: 960 bytes)
```

Method pre-computation (`@eval`):

Julia 0.6.4
```julia
julia> for i in 1:8
           @btime FixedDecimal{Int,$i}.(1:100)
       end
  51.640 μs (35 allocations: 2.31 KiB)
  51.611 μs (35 allocations: 2.31 KiB)
  53.136 μs (35 allocations: 2.31 KiB)
  52.384 μs (35 allocations: 2.31 KiB)
  52.791 μs (35 allocations: 2.31 KiB)
  53.087 μs (35 allocations: 2.31 KiB)
  53.697 μs (35 allocations: 2.31 KiB)
  53.469 μs (35 allocations: 2.31 KiB)
```

Julia 1.0.1
```julia
julia> for i in 1:8
           @btime FixedDecimal{Int,$i}.(1:100)
       end
  375.854 ns (3 allocations: 960 bytes)
  416.121 ns (3 allocations: 960 bytes)
  1.977 μs (3 allocations: 960 bytes)
  511.617 ns (3 allocations: 960 bytes)
  2.072 μs (3 allocations: 960 bytes)
  1.938 μs (3 allocations: 960 bytes)
  2.299 μs (3 allocations: 960 bytes)
  507.642 ns (3 allocations: 960 bytes)
```

Julia master
```julia
julia> for i in 1:8
           @btime FixedDecimal{Int,$i}.(1:100)
       end
  377.093 ns (3 allocations: 960 bytes)
  421.030 ns (3 allocations: 960 bytes)
  774.255 ns (3 allocations: 960 bytes)
  510.917 ns (3 allocations: 960 bytes)
  1.498 μs (3 allocations: 960 bytes)
  2.077 μs (3 allocations: 960 bytes)
  814.361 ns (3 allocations: 960 bytes)
  525.984 ns (3 allocations: 960 bytes)
```

Conclusion: Julia 0.6 is beyond help ;). I think the timing differences between the function barrier and the pre-computing is just noise. I'll update to use @KristofferC's approach as this avoids method pollution.

omus:
Any further comments? I'll squash and merge in an hour from now.

NHDaly:
> @code_lowered doesn't really tell you anything, use @code_typed.

Oops, duh right sorry. Still, no change. The point is, unless you remove the `applicable` call from `max_exp10`, you won't see any performance improvement.

NHDaly:
Ah, haha sorry, i just missed your deadline. I was about to say:
Actually, i don't think the `throw` function barrier does anything at all.

I think all the performance improvement simply comes from deleting the `applicable` call, and adding the manual definition for `BigInt`.

PR #34, opened 2018-10-26
Fix constructor performance by swapping runtime-evaluated `applicable` with static `isbitstype`

NHDaly:
Okay!

I actually think this is probably the best fix, just because it's the simplest. Basically, the whole problem stemmed from the call to `applicable`, which prevents const-folding since its value depends on global state. The return value of `applicable` can change if you define new methods, so it has to check every time it would be called at runtime.

I just switched it out for `isbitstype`, which I think is a good enough proxy, is closer to what we actually meant with this check anyway, and of course depends only on the type definition, so its result is static. This allows the entire method to be compiled away.

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -27,7 +27,7 @@ module FixedPointDecimals
 
 export FixedDecimal, RoundThrows
 
-using Compat: lastindex, something
+using Compat: lastindex, something, isbitstype
 
 import Compat: floatmin, floatmax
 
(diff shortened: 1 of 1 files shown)


NHDaly:
(oh, `isbitstype` isn't defined in julia 0.6, thus the build failure. maybe there's an 0.6 alternative?)

KristofferC:
I think it is just `isbits` in 0.6.

KristofferC:
Isn't it surprising that defining

```
mutable struct IntWrapper
    x::Int
end
```

and

```
struct IntWrapper2
    x::Int
end
```

with the exact same definitions on them would give different answer here. I don't see how `isbitstype` (which is about layout of data) has anything to do with the documented behavior of this function.

NHDaly:
Ah, yeah, you're probably right.. :/ that's too bad. I only thought of `isbitstype` because it seemed to be the thing distinguishing `BigInt` from the other `Integer` types. But I suppose you're right that the above distinction is surprising.

PR #36, opened 2018-11-02
Fix construction with same type performing convert

NHDaly:
Before this PR, constructing a new FD with the same type called `convert`. That happened due to this line, since `FD<: Real`:
https://github.com/JuliaMath/FixedPointDecimals.jl/blob/cc6aed31ec84e8e6e0f428f9a36d34652b48a9c2/src/FixedPointDecimals.jl#L113

```
julia> f =  FixedPointDecimals.FixedDecimal{Int8, 2}(1)
>> FixedDecimal{Int8,2}(1.00)

julia> @which typeof(f)(f)
>> (::Type{T})(x::Real) where T<:FixedDecimal in FixedPointDecimals at /Users/daly/.julia/dev/FixedPointDecimals/src/FixedPointDecimals.jl:113
```

This PR adds a "copy constructor" to avoid that. I'm not really sure what the idiomatic name for it would be in julia, "copy constructor" of course references the name of the same concept in C++. :)

Perhaps a separate issue is that for some reason `convert` is super expensive when `T=Int128`, which is how I stumbled across this problem. But either way, i think this PR could make sense.

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -99,6 +99,9 @@ struct FixedDecimal{T <: Integer, f} <: Real
             _throw_storage_error(f, T, n)
         end
     end
+
+    # Copy constructor -- prevents unneeded work to convert between the same types
+    FixedDecimal{T,f}(x::FixedDecimal{T,f}) where {T<:Integer,f} = new{T,f}(x.i)
 end
 
 @noinline function _throw_storage_error(f, T, n)


NHDaly:
> Can you post some performance benchmarks?

Yes, definitely!

Sorry, i had this tiny "benchmark" in the commit text, but didn't post it in the PR:

    Before this change:
    ```
      julia> f = FixedDecimal{Int128,2}(3.25)
      >> FixedDecimal{Int128,2}(3.25)

      julia> @btime x = FixedDecimal{Int128, 2}($f)
        7.272 μs (159 allocations: 2.73 KiB)
      >> FixedDecimal{Int128,2}(3.25)
    ```

    After this change:
    ```
      julia> f = FixedDecimal{Int128,2}(3.25)
      >> FixedDecimal{Int128,2}(3.25)

      julia> @btime x = FixedDecimal{Int128, 2}($f)
        1.508 ns (0 allocations: 0 bytes)
      >> FixedDecimal{Int128,2}(3.25)
    ```

But I'll also post up a better benchmark in a second.

NHDaly:
Okay, so here's the benchmarks that originally led me here.

Here's the benchmark in a gist:
https://gist.github.com/NHDaly/a8fae0d1d65ab1066c585c27e54146fa

And the results in a google spreadsheet:
https://docs.google.com/spreadsheets/d/1Lc3ughgwwK25cpwbnuLRxw19EtaMxdCsMmtnspT_4H8/edit?usp=sharing

For some reason, even just element-wise _copying_ an array of `FixedDecimal{Int128, 2}` into another array, allocates like crazy. (14,000,000 allocations / 1,000,000 elements).

I was trying to find the source of that allocation, which led me to this change. It doesn't actually address the fact that Int128s are allocating when copied, when the other sizes don't, but it _does_ eliminate unnecessary work, by preventing a `convert` call when the types are identical.

Separately, though, it would be really nice to identify the source of the allocations for Int128.

omus:
Reposting results:

|              |              | *Operation*   | *Values*   |               |            |               |            |               |            |               |            |
|:-------------|:-------------|:--------------|:-----------|:--------------|:-----------|:--------------|:-----------|:--------------|:-----------|:--------------|:-----------|
|              |              | **identity**  |            | **÷**         |            | **+**         |            | **/**         |            | __*__         |            |
| *Category*   | *Type*       | **time (ms)** | **allocs** | **time (ms)** | **allocs** | **time (ms)** | **allocs** | **time (ms)** | **allocs** | **time (ms)** | **allocs** |
| Int          | Int32        | 1.35          | 0          | 5.16          | 0          | 1.47          | 0          | 2.35          | 0          | 1.61          | 0          |
|              | Int64        | 1.86          | 0          | 18.66         | 0          | 1.89          | 0          | 2.46          | 0          | 1.95          | 0          |
|              | Int128       | 3.77          | 0          | 26.07         | 0          | 3.85          | 0          | 16.74         | 0          | 3.95          | 0          |
| Float        | Float32      | 1.35          | 0          | 28.97         | 0          | 1.47          | 0          | 1.75          | 0          | 1.47          | 0          |
|              | Float64      | 1.85          | 0          | 27.37         | 0          | 1.88          | 0          | 2.45          | 0          | 1.89          | 0          |
| FixedDecimal | FD{Int32,2}  | 1.35          | 0          | 5.16          | 0          | 1.48          | 0          | 38.20         | 0          | 31.73         | 0          |
|              | FD{Int64,2}  | 1.86          | 0          | 18.75         | 0          | 1.89          | 0          | 59.18         | 0          | 47.03         | 0          |
|              | FD{Int128,2} | 1320.01       | 14000000   | 26.19         | 0          | 1324.35       | 14000000   | 7267.32       | 72879639   | 6139.13       | 62000000   |

NHDaly:
(Note that after this PR, the benchmarks don't change. This is more like a minor optimization I discovered while researching the weird allocations. Maybe I'll open another issue for the benchmarks then separately!) :)

NHDaly:
(Opened #37 to discuss the weird allocation and other questions)

Issue #37, opened 2018-11-05
Performance investigation


NHDaly:
Splitting out the benchmark discussion from https://github.com/JuliaMath/FixedPointDecimals.jl/pull/36.

------

Here's the benchmark in a gist:
https://gist.github.com/NHDaly/a8fae0d1d65ab1066c585c27e54146fa

And the results in a google spreadsheet:
https://docs.google.com/spreadsheets/d/1Lc3ughgwwK25cpwbnuLRxw19EtaMxdCsMmtnspT_4H8/edit?usp=sharing

-------

Here are the results:

  |   | Operation | Values |   |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
  |   | identity |   | ÷ |   | + |   | / |   | * |  
Category | Type | time (ms) | allocs | time (ms) | allocs | time (ms) | allocs | time (ms) | allocs | time (ms) | allocs
Int | Int32 | 1.35 | 0 | 5.16 | 0 | 1.47 | 0 | 2.35 | 0 | 1.61 | 0
  | Int64 | 1.86 | 0 | 18.66 | 0 | 1.89 | 0 | 2.46 | 0 | 1.95 | 0
  | Int128 | 3.77 | 0 | 26.07 | 0 | 3.85 | 0 | 16.74 | 0 | 3.95 | 0
Float | Float32 | 1.35 | 0 | 28.97 | 0 | 1.47 | 0 | 1.75 | 0 | 1.47 | 0
  | Float64 | 1.85 | 0 | 27.37 | 0 | 1.88 | 0 | 2.45 | 0 | 1.89 | 0
FixedDecimal | FD{Int32,2} | 1.35 | 0 | 5.16 | 0 | 1.48 | 0 | 38.20 | 0 | 31.73 | 0
  | FD{Int64,2} | 1.86 | 0 | 18.75 | 0 | 1.89 | 0 | 59.18 | 0 | 47.03 | 0
  | FD{Int128,2} | 1320.01 | 14000000 | 26.19 | 0 | 1324.35 | 14000000 | 7267.32 | 72879639 | 6139.13 | 62000000

--------------

Here are my current question:

- [ ] For some reason, even just element-wise copying an array of `FixedDecimal{Int128, 2}` into another array, allocates like crazy. (14,000,000 allocations / 1,000,000 elements).
     - Where do these allocations come from?
[...]

PR #38, opened 2018-11-07
Performance fix: avoid unnecessary `convert` on setindex!、copy when type is same.

NHDaly:
Add `convert(::Type{T}, x::T) where {T<:FD} = x`

This fixes unnecessary copies on `setindex!`, etc.

Fixes the allocations issue for `identity` from this table, by preventing the widening to `BigInt`:
https://github.com/JuliaMath/FixedPointDecimals.jl/issues/37

This is the same situation as this issue:
JuliaLang/julia#17559

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -274,6 +274,8 @@ function round(::Type{FD{T, f}}, x::Rational, ::RoundingMode{:Nearest}=RoundNear
 end
 
 # conversions and promotions
+convert(::Type{FD{T, f}}, x::FD{T, f}) where {T, f} = x  # Converting an FD to itself is a no-op
+
 function convert(::Type{FD{T, f}}, x::Integer) where {T, f}
     reinterpret(FD{T, f}, T(widemul(x, coefficient(FD{T, f}))))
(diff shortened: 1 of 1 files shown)


src/FixedPointDecimals.jl
```diff
@@ -274,6 +274,8 @@ function round(::Type{FD{T, f}}, x::Rational, ::RoundingMode{:Nearest}=RoundNear
 end
 
 # conversions and promotions
+convert(::Type{T}, x::T) where {T<:FD} = x  # Converting an FD to itself is a no-op
```
omus: If you modify this to:
```julia
convert(::Type{FD{T, f}}, x::FD{T, f}) where {T, f} = x
```
Julia 0.6 will be happy.

NHDaly: ✔️ Thanks Curtis! I had that at first, but then i thought that this version was simpler.

NHDaly: Done.

NHDaly:
(I've also removed the copy constructor that I added in #36, since this is the more general and correct fix.)

PR #40, opened 2018-11-29
Make FD÷FD return an FD.

NHDaly:
Before this, `FD{T,f} ÷ FD{T,f}` returned `T`, now it returns `FD{T,f}`.

Same for `div`, `fld`, and `fld`.

--------------

Here's an example of the old behavior. Addition and division for two `FixedDecimals` returns a `FixedDecimal`, but integer division (`div`) was returning the raw storage type `T` instead of wrapping it in a FixedDecimal.

```julia
julia> FixedDecimal{Int32, 2}(2)+FixedDecimal{Int32, 2}(1)
>> FixedDecimal{Int32,2}(3.00)

julia> FixedDecimal{Int32,2}(4.00) / FixedDecimal{Int32,2}(2.00)
>> FixedDecimal{Int32,2}(2.00)

julia> FixedDecimal{Int32, 2}(2)÷FixedDecimal{Int32, 2}(1)
>> 2
```

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -512,19 +512,19 @@ end
 @testset "truncating div" begin
     @testset "div by 1" begin
         @testset for x in keyvalues[FD2]
-            @test x ÷ one(x) == trunc(x)
+            @test x ÷ one(x) === trunc(x)
 
             # signed integers using two's complement have one additional negative value
-            if x < 0 && trunc(x) == typemin(x)
+            if x < 0 && trunc(x) === typemin(x)
                 @test_throws InexactError x ÷ -one(x)
             else
-                @test x ÷ -one(x) == -trunc(x)
+                @test x ÷ -one(x) === -trunc(x)
             end
         end
     end
     @testset "div by 2" begin
         @testset for x in keyvalues[FD2]
-            @test x ÷ 2one(x) == x ÷ 2 == x.i ÷ FD2(2).i
+            @test x ÷ 2one(x) === x ÷ 2 === FD2(x.i ÷ FD2(2).i)
         end
     end
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -509,6 +509,26 @@ end
     end
 end
 
+@testset "truncating div" begin
+    @testset "div by 1" begin
+        @testset for x in keyvalues[FD2]
+            @test x ÷ one(x) == trunc(x)
+
+            # signed integers using two's complement have one additional negative value
+            if x < 0 && trunc(x) == typemin(x)
+                @test_throws InexactError x ÷ -one(x)
+            else
+                @test x ÷ -one(x) == -trunc(x)
+            end
+        end
+    end
+    @testset "div by 2" begin
+        @testset for x in keyvalues[FD2]
+            @test x ÷ 2one(x) == x ÷ 2 == x.i ÷ FD2(2).i
+        end
+    end
+end
+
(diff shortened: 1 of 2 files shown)


NHDaly:
Also, I noticed that there weren't any unit tests for `div`, so I added a few simple ones.

omus:
Since `/` returns a FD I do think it makes sense to do the same for `div`. We probably do need to have some of the widening logic from `/` here as well though.

NHDaly:
> One advantage of returning `T` is that there's no danger of overflow, whereas when returning the FD there is. Is there a reason why the FD is desirable here?

Yeah, this is a good point, @TotalVerb. By returning a `T` right now, we avoid overflows in places where it would happen otherwise, e.g.:
```julia
julia> FixedPointDecimals.FixedDecimal{Int16, 4}(1) ÷ FixedPointDecimals.FixedDecimal{Int16, 4}(0.0005)
>> 2000

julia> FixedPointDecimals.FixedDecimal{Int16, 4}(2000)
ERROR: InexactError: trunc(Int16, 20000000)
```

That said, I agree with `omus` that since all the other operations (such as `/`) return an FD, I think it makes sense here as well. For example, if `+` returned an integer it, too, could avoid overflows but that wouldn't really make sense. That said, this one _is_ a bit different than `+` since it has to do a bit of work to convert the result _back_ to a FD when the actual operation itself causes you to produce the integer representation... But I dunno, it just seems weird for users to need to remember to convert back to an FD every time they integer divide them. And if the user really wants to get the `int` representation back, they can always just divide their internal values: `x.i ÷ y.i`?


--------------

> @omus commented 3 days ago
> We probably do need to have some of the widening logic from `/` here as well though.

Actually, I don't think we do, although it would probably be worth adding a comment explaining why we don't.
When you divide the two FixedDecimals, is you automatically eliminate all of the fractional part of the answer (`FD{Int,2}(3) ÷ FD{Int,2}(2)` --> `300÷200` == `1`). 

`/` needs to widen the inputs and multiply by `10^f` ahead of time, specifically to _avoid_ eliminating the fractional part of the answer, since we don't want to lose precision. But in the case of truncating integer division (`÷`), "eliminating the fractional part of the answer" is exactly what we want! So it should be enough as is, I think.

NHDaly:
Added a comment here to explain why we don't need the widening logic. Does this look okay?:

https://github.com/JuliaMath/FixedPointDecimals.jl/blob/becc3dfc5d4830052a2b5951ce4cc3d7b629943f/src/FixedPointDecimals.jl#L307-L311

test/runtests.jl
```diff
@@ -509,6 +509,26 @@ end
...
+                @test x ÷ -one(x) == -trunc(x)
+            end
+        end
+    end
+    @testset "div by 2" begin
+        @testset for x in keyvalues[FD2]
+            @test x ÷ 2one(x) == x ÷ 2 == x.i ÷ FD2(2).i
+        end
+    end
+end
```
omus: Tests should check the output type:
```julia
julia> FD2 = FixedDecimal{Int,2}; FD(1) ÷ FD(1) == 1
true
```

NHDaly: Thanks, great point. I've updated the tests to all be `===` which I think should be sufficient. Notably, the tests now fail without the changes in this PR.

Let me know if you'd prefer anything else! :)

NHDaly: (I've also added a section that explicitly tests the return type)

PR #42, opened 2018-12-06
Add Decimal Representations Comparison benchmark.

NHDaly:
Opening this PR to discuss merging benchmarks into this repo, so that we can track performance across commits/versions.

I'm not sure if there's a usual structure to follow for putting benchmarks into a julia repo? Do other repos besides the main Julia repo use Nanosoldier?

Also, in its current form, the benchmark compares against raw `Int` and `Float` types, but for all the operations except division, those types execute the operation in a single clock tick, so it's almost not worth spending the computation to measure them... So maybe we can simplify this file to just measure FixedDecimals.

Anyway, looking forward to figuring out the best way to do this with you! :)

### DIFF
--- a/bench/decimal-representation-comparisons.jl
+++ b/bench/decimal-representation-comparisons.jl
@@ -1,0 +1,122 @@
+# Decimal Representation Comparisons
+#
+# This benchmark compares the performance of several numeric representations, over various
+# numeric operations (+,-,*,/,÷...) on large arrays of numbers, in order to guide
+# decision-making about how to represent fixed-decimal numbers.
+#
+# It compares fixed-decimal types against the builtin Int and Float types of various sizes.
+# The output is written to a .csv file in the same directory as this file.
+
+module DecimalRepresentationComparisons
+
+using FixedPointDecimals
+using Random
+using BenchmarkTools, Statistics
+using DataFrames
+using CSV
+
+decimal_precision = 2
+
+# Express that data through the various types. Round it for integers.
+fd_FixedPointDecimal_types = [
+    FixedPointDecimals.FixedDecimal{Int32, decimal_precision},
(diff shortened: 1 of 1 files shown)


TotalVerb:
I think this makes sense. We do have the benchmarks in bench/ on JSON.jl also: https://github.com/JuliaIO/JSON.jl/tree/master/bench, so there's precedent.

NHDaly:
@omus that PkgBenchmark seems nice, thanks for the link. (I'm sending a couple PRs now to clean it up for 1.0 so we can tag a version there. 😄)

I guess do you want me to play with setting that up before merging this PR in? I think that seems reasonable

NHDaly:
So I just want to leave a status update here.

So I think this basically works. The benchmarks run (and after the merged-changes in https://github.com/JuliaCI/PkgBenchmark.jl/pull/75, they should correctly and precisely be measuring _only_ the time for each operation (not copying the value, reading from an array, etc)).

The remaining blocker to merging is that it's _extremely_ variable, so much so that I don't think it's useful. Even when running on a single computer, comparing the a single commit against itself, `PkgBenchmark` consistently reports statistically significant variance. If anyone has any advice about how to diagnose this, that would be much appreciated!

I've tried several things trying to pinpoint the source of the variance, but haven't had any luck:
- I tried simplifying the benchmarks to just be `@benchmarkable $op($x, $x)`, but saw the same variance there.
- I tried statically compiling a sysimg containing FixedPointDecimals, and using that for running the benchmarks, which didn't help.
- One of my coworkers tried disabling the GC (i'm not sure what steps they took), but said that didn't help either.

Does anyone have any other ideas? Without this, this seems not very useful. Sometimes the swings are as large as 100% or 200%, so I'm not sure we'd get meaningful feedback on PRs.

PR #29, opened 2018-10-16
Fix travis coverage on 1.0

TotalVerb:
I think there is a problem with codecov on Julia 1.0 due to the Pkg stdlib change. Let's see if this fixes it.

PR #44, opened 2018-12-18
Update .travis.yml coverage lines for julia 1.0

NHDaly:
- Add `using Pkg` if `VERSION` >= 0.7.
- The default Travis scripts now keep the `cwd` inside the package, so we don't need the `cd()` line.

We can submit this instead of https://github.com/JuliaMath/FixedPointDecimals.jl/pull/29.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,6 @@ julia:
   - nightly
 notifications:
   email: false
-# uncomment the following lines to override the default test script
-#script:
-#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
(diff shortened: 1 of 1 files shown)


NHDaly:
I based this off of the Example.jl repo here: https://github.com/JuliaLang/Example.jl/blob/master/.travis.yml

omus:
Looking at the coverage I'm not sure we're [actually submitting coverage for Julia 0.6](https://codecov.io/gh/JuliaMath/FixedPointDecimals.jl/src/0787a9f25fce46ea954b3d799c9ebec9a6e3835c/src/FixedPointDecimals.jl#L230). Since Julia 0.6 is now unmaintained I'm fine with dropping support here.

PR #45, opened 2018-12-19
Speed up * for FD{Int64} via custom `fldmod_by_const`

NHDaly:
Add a custom implementation of `fldmod(x::T,y)` for when `y` is a const,
and call it whenever T >= Int128.

This improves performance for `*(a::FD{Int64}, b::FD{Int64})`, b/c `*` passes
the result of `widemul(a,b)` into `fldmod` -- and LLVM isn't able to
optimize `fldmodinline(::Int128,::Const{Int128})` nearly as well as we'd like.

Here, we opt to use a custom implementation of `fldmod` in that case,
which runs significantly faster, making the multiply faster.

--------

Specifically, the custom `fldmod_by_const` implementation is based off of the assembly generated by LLVM for `fld(x,100)`. One big reason it can't emit similarly optimized code when `x` is an `Int128` is because there isn't a single `op` to efficiently get the upper-half of the "split-widemul". This requires adding a function that does it, which I've called `splitmul_upper`.

--------

Note that this PR only directly improves performance for `FD{Int64}`: `FD{Int128}` still widens to `BigInt`, and we do not use the custom implementation in that case.

I'm following this up with https://github.com/RelationalAI-oss/FixedPointDecimals.jl/pull/7, which changes `Int128` to widen to a 256-bit integer implementation, so that we can use the custom `fldmod` for `FD{Int128}` as well! :)


NHDaly:
Unfortunately I still haven't found a good solution for the automated benchmarks (i'm working on it!), but in the meantime, i've re-run the benchmark locally. Here's the updated timings for just the FixedDecimal multiplication:

  | time (ns) | allocs
-- | -- | --
FD{ Int32,2} | 2.09 | 0
FD{ Int64,2} | 8.25 | 0
FD{Int128,2} | 2186.16 | 31000069

Compared to master (after merging https://github.com/JuliaMath/FixedPointDecimals.jl/pull/43), this PR reduces the time for `FD{Int64,2}` by 50%, down from `16.61 ns` here: https://github.com/JuliaMath/FixedPointDecimals.jl/pull/43#issue-236631643! :)

test/fldmod_by_const.jl
```diff
@@ -0,0 +1,48 @@
+using Test
+using FixedPointDecimals
+
+@testset "splitmul_upper" begin
+    vals = [0xA3D70A3D70A3D70B, 0x6666666666666666, typemax(Int16), typemax(Int16),
+            typemax(Int64), typemax(Int64)]
+    @testset "$a" for a in vals
+        @testset "$b" for b in vals
+            @testset "$f" for f in (unsigned, signed)
```
omus: Having all of these as separate testsets seems noisy to me. Consider doing:
```julia
@testset "..." for (a, b, f) in Iterators.product(vals, vals, (unsigned, signed))
    ...
end
```

src/fldmod_by_const.jl
```diff
@@ -0,0 +1,156 @@
...
+    toshift = _leading_zeros(invcoeff)
+    invcoeff = invcoeff << toshift
+    # Now, truncate the result to sizeof(T), rounding to maintain precision.
+    invcoeff = _round_to_even(fldmod(invcoeff, typemax(UT))..., typemax(UT)) % T
+    return invcoeff, toshift
+end
+# These are needed to handle Int128, which widens to BigInt, since BigInt doesn't have typemax
+twoToTheSizeOf(::Type{T}) where {T} = typemax(widen(unsigned(T)))
+twoToTheSizeOf(::Union{Type{Int128}, Type{UInt128}}) = BigInt(2)^256
+twoToTheSizeOf(::Type{BigInt}) = BigInt(2)^256
```
omus: Stick to underscores

src/fldmod_by_const.jl
```diff
@@ -0,0 +1,156 @@
...
+    halfbits = nbits(al)
+    # /* compute partial products */
+    p0 = widemul(al, bl);
+    p1 = widemul(al, bh);
+    p2 = widemul(ah, bl);
+    p3 = widemul(ah, bh);
+    # /* sum partial products */
+    carry = ((p0 >> halfbits) + (p1%halfT) + (p2%halfT)) >> halfbits;
+    return p3 + (p2 >> halfbits) + (p1 >> halfbits) + carry;
+end
```
omus: You may as well make the comments and remove the `;`

src/fldmod_by_const.jl
```diff
@@ -0,0 +1,156 @@
...
+narrow(::Type{Int128}) = Int64
+narrow(::Type{Int64}) = Int32
+narrow(::Type{Int32}) = Int16
+narrow(::Type{Int16}) = Int8
+narrow(::Type{UInt128}) = UInt64
+narrow(::Type{UInt64}) = UInt32
+narrow(::Type{UInt32}) = UInt16
+narrow(::Type{UInt16}) = UInt8
+
+"""    nbits(x) = sizeof(x)*8 """
```
omus: There really isn't a point to this docstring

src/FixedPointDecimals.jl
```diff
@@ -165,10 +167,20 @@ _round_to_even(q, r, d) = _round_to_even(promote(q, r, d)...)
 # correctness test suite.
 function *(x::FD{T, f}, y::FD{T, f}) where {T, f}
     powt = coefficient(FD{T, f})
-    quotient, remainder = fldmodinline(widemul(x.i, y.i), powt)
+    quotient, remainder = custom_fldmod(widemul(x.i, y.i), powt)
     reinterpret(FD{T, f}, _round_to_even(quotient, remainder, powt))
 end
 
+@inline function custom_fldmod(x::T,y) where T
```
omus: `_fldmod` is fine for the name. Should have space after `,`

src/fldmod_by_const.jl
```diff
@@ -0,0 +1,156 @@
...
+
+Returns `fldmod(x,C)`, implemented efficiently when `C` is a (positive) static Const.
+
+REQUIRES:
+  - `C` must be greater than `0`
+"""
+function fldmod_by_const(x, y::Val{C}) where {C}
+    d = fld_by_const(x, y)
+    return d, manual_mod(promote(x, C, d)...)
+end
```
omus: Try to keep these methods together

src/FixedPointDecimals.jl
omus: Actually since `fldmodinline` is also defined by this package why not make that the name here too? I'd also probably move this into your separate file

test/fldmod_by_const.jl
```diff
@@ -0,0 +1,48 @@
+using Test
```
omus: These tests aren't actually run by `runtests.jl`

NHDaly:
In this PR, we used the following check so that we'll only use the custom implementation of `fldmod` when it's needed, because LLVM can't do it itself:
https://github.com/JuliaMath/FixedPointDecimals.jl/pull/45/files#diff-ea4e0198fdaaab3db0fa802fdb33c161R178

It might look a little weird as it's currently phrased, because only `FD{Int64}` would use it (when the Int64 widens to Int128, then we call `fldmod(::Int128, 100)`). In this PR, `FD{Int128}` doesn't get funneled into the `fldmod_by_const`, because Int128 widens to `BigInt`.

In the next PR (https://github.com/RelationalAI-oss/FixedPointDecimals.jl/pull/7), though, `FD{Int128}` uses the same implementation as well, because it widens to `Int256`, and then we call `fldmod_by_const(Int256)`. Here are the performance results from that change: https://github.com/RelationalAI-oss/FixedPointDecimals.jl/pull/7#issuecomment-448774559.

src/FixedPointDecimals.jl
NHDaly: I'm down to combine those two functions, but the one thing I don't like about the name `fldmodinline` applied to `custom_fldmod` is that `fldmodinline` implies it can be a reasonable drop-in for `fldmod` anytime. But in fact, calling `custom_fldmod` with a _non-const_ `y` (such as in `/(::FD,::FD)`) would be terrible because creating the `Val` with runtime data would be super expensive.

Actually, i think a good name for this function would be `fldmod_by_const`, but I was using that for the name of the actual manual implementation when y is a `Val`... Should I just make the `::Val` case a separate _method_ with the same name? Or do you have other ideas? Thanks!

test/fldmod_by_const.jl
```diff
@@ -0,0 +1,48 @@
...
+                a,b = promote(f(a),f(b))
+                @test (FixedPointDecimals.splitmul_upper(a,b) ==
+                        FixedPointDecimals.splitint(widemul(a,b))[1])
+            end
+        end
+    end
+end
+
+@testset "div_by_const" begin
+    vals = [2432, 100, 0x1, Int32(10000), typemax(Int64), typemax(Int16)]
```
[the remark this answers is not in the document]
NHDaly: Happily! Done.

You were right of course. I've added a big note about fixing this above!

src/FixedPointDecimals.jl
NHDaly: I've just pushed up a commit (https://github.com/JuliaMath/FixedPointDecimals.jl/pull/45/commits/33e11d0c5ae9356b14b34d2c34e4f276b1494bfa) which does this.

I combined those two functions, and moved it to the separate fldmod_by_const.jl file. I called it `fldmod_by_const`, since that's the whole point of it! :) In the case where we can rely on LLVM to correctly do the dividing by a const, we do that! And when we can't we use the implementation taking a `::Val{C}`.

I think this is much cleaner. Thanks! :)

src/fldmod_by_const.jl
NHDaly: Yeah agreed, they were a bit all-over the place before.

I just reordered the whole file in this commit (https://github.com/JuliaMath/FixedPointDecimals.jl/pull/45/commits/f1ff3a8dd5ba530c01fbdff07810c3eba813a5cc) to be much cleaner I think.

It starts with the `fldmod_by_const`, then goes top-down more or less depth-first, so that each new function is defined right after it is first referenced, and the methods are kept together. Hopefully this is easier to read. Let me know what you think!

NHDaly:
Happy New Year you all! :) Hope you had a nice holiday. :)


I've cleaned this up a bit after some time away and I think i've addressed all the outstanding comments! :) Please take another look when you get time.

Cheers!

omus:
I'll try to re-review this week. I'll note the CI is failing on Julia 0.6.

src/fldmod_by_const.jl
```diff
@@ -0,0 +1,188 @@
...
+
+narrow(::Type{Int128}) = Int64
+narrow(::Type{Int64}) = Int32
+narrow(::Type{Int32}) = Int16
+narrow(::Type{Int16}) = Int8
+narrow(::Type{UInt128}) = UInt64
+narrow(::Type{UInt64}) = UInt32
+narrow(::Type{UInt32}) = UInt16
+narrow(::Type{UInt16}) = UInt8
```
omus: Add `narrow(::Type{Int8}) = Int8` and make sure to add tests for this.

src/fldmod_by_const.jl
```diff
@@ -0,0 +1,188 @@
...
+@inline function splitmul_upper(a::T, b::T) where T<:Signed
+    uresult = unsigned_splitmul_upper(unsigned(a), unsigned(b))
+    return signed(uresult) - ((a < 0) ? b : 0) - ((b < 0) ? a : 0)
+end
+@inline splitmul_upper(a, b) = splitmul_upper(promote(a, b)...)
+
+# Implemenation based on umul32hi, from https://stackoverflow.com/a/22847373/751061
+# Compute the upper half of the widened product of two unsigned integers.
+# Example: `widemul(0x0020,0x2002) == 0x0004_0040` vs
+#          `unsigned_splitmul_upper(0x0020,0x2002) == 0x0004`
```
omus: Comment should go into the purpose of this function. I found the link to be more insightful than this comment.

src/fldmod_by_const.jl
```diff
@@ -0,0 +1,188 @@
...
+function fldmod_by_const(x, y::Val{C}) where {C}
+    d = fld_by_const(x, y)
+    return d, manual_mod(promote(x, C, d)...)
+end
+
+# Calculate fld(x,y) when y is a Val constant.
+# The implementation for fld_by_const was lifted directly from Base.fld(x,y), except that
+# it uses `div_by_const` instead of `div`.
+fld_by_const(x::T, y::Val{C}) where {T<:Unsigned, C} = div_by_const(x, y)
+function fld_by_const(x::T, y::Val{C}) where {T<:Integer, C}
```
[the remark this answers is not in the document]
NHDaly: Hmm, do you think so? I didn't want to prevent anyone from using a custom integer type, such as, for example, a custom `Int256` if they were feeling crazy. Or like maybe some has like an `Int20` type or something for some reason.


But I do agree this optimization is pretty specific to the bit layout of the integer types, so it's reasonable to want to limit it to "any integer type," because i can imagine other integer types (`BigInt`) that it wouldn't work for.

What do you think?

src/fldmod_by_const.jl
```diff
@@ -0,0 +1,188 @@
...
+REQUIRES:
+  - `C` _must be_ greater than `0`
+"""
+function div_by_const(x::T, ::Val{C}) where {T, C}
+    # These checks will be compiled away during specialization.
+    if C == 1
+        return x
+    elseif ispow2(C)
+        return div(x, C)
+    elseif (C <= 0)
```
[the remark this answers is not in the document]
NHDaly: Done. Thanks.

src/fldmod_by_const.jl
```diff
@@ -0,0 +1,188 @@
...
+# These are needed to handle Int128, which widens to BigInt, since BigInt doesn't have typemax
+two_to_the_size_of(::Type{T}) where {T} = typemax(unsigned(T))
+two_to_the_size_of(::Type{BigInt}) = BigInt(2)^256
+
+# This special-purpose leading_zeros is needed to handle Int128, which widens to BigInt
+_leading_zeros(x) = leading_zeros(x)
+# BigInt doesn't have a concept of "leading zeros", but since we _know_ the value being
+# passed here will fit in 256-bits (per two_to_the_size_of), we can pretend this is a
+# 256-bit integer, take just the upper half as a UInt128, and count the leading zeros there.
+_leading_zeros(x::BigInt) = leading_zeros((x >> 128) % UInt128)
```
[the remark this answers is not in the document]
NHDaly: Hmm, maybe this is a case where control-flow is preferable to dispatch, and i should just replace the call to `_leading_zeros(invcoeff)` with a check like this:
```julia
if invcoeff isa BigInt
    # BigInt doesn't have a concept of "leading zeros", but since we _know_ the value being
    # passed here will fit in 256-bits (per two_to_the_size_of), we can pretend this is a
    # 256-bit integer, take just the upper half as a UInt128, and count the leading zeros there.
    leading_zeros((x >> 128) % UInt128)
else
    leading_zeros(x)
end
```
? What do you think about that? I think that's actually decent because it prevents this function from being misused in the future as you rightly point out.

If you agree, I'll make this change!

src/fldmod_by_const.jl
```diff
@@ -0,0 +1,188 @@
...
+
+
+@inline function splitmul_upper(a::T, b::T) where T<:Unsigned
+    return unsigned_splitmul_upper(a, b)
+end
+@inline function splitmul_upper(a::T, b::T) where T<:Signed
+    uresult = unsigned_splitmul_upper(unsigned(a), unsigned(b))
+    return signed(uresult) - ((a < 0) ? b : 0) - ((b < 0) ? a : 0)
+end
+@inline splitmul_upper(a, b) = splitmul_upper(promote(a, b)...)
```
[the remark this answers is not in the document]
NHDaly: Okay great, agreed! Done.

src/fldmod_by_const.jl
```diff
@@ -0,0 +1,188 @@
...
+
+# Implemenation based on umul32hi, from https://stackoverflow.com/a/22847373/751061
+# Compute the upper half of the widened product of two unsigned integers.
+# Example: `widemul(0x0020,0x2002) == 0x0004_0040` vs
+#          `unsigned_splitmul_upper(0x0020,0x2002) == 0x0004`
+@inline function unsigned_splitmul_upper(a::T, b::T) where T<:Unsigned
```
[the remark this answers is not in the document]
NHDaly: I think I did that just to really _emphasize_ that this only works for unsigned integers, and to make it clearer that the _signed_ version is calling the unsigned one and then doing some bit twiddling. But I'll change it since I agree that should be obvious enough I think.

It might've just been carried over from the C implementation without me thinking about it.. Fixed, thanks.

src/fldmod_by_const.jl
```diff
@@ -0,0 +1,188 @@
...
+    # Split operands into halves
+    ah, al = splitint(a)
+    bh, bl = splitint(b)
+    halfT = typeof(ah)
+    halfbits = nbits(al)
+    # Compute partial products
+    p0 = widemul(al, bl);
+    p1 = widemul(al, bh);
+    p2 = widemul(ah, bl);
+    p3 = widemul(ah, bh);
```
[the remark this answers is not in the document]
NHDaly: Done.

src/fldmod_by_const.jl
NHDaly: :/ unless i make some other changes, I don't think that's a good idea.

Maybe I should change the name of this to not be `narrow`, because what I'm really relying on is that `narrow(T)` will be exactly **half** the size of `T`, so that `splitint` will split an integer into the upper and lower halves.

I like the fact that as is, `splitint(Int8(3))` throws `MethodError: no method matching narrow(::Type{Int8})`, because yeah, you _can't_ split it into two halves since there's no smaller type to hold the result! But maybe if this isn't clear I need to either rename `narrow` or add better comments?

src/fldmod_by_const.jl
NHDaly: Thanks, this was good feedback.

I've hopefully improved the comments in this last commit: https://github.com/JuliaMath/FixedPointDecimals.jl/pull/45/commits/dd78a51d71d51dff687bcf00192c13878b654357

NHDaly:
(note: i've Resolved the threads I've addressed. Any remaining unresolved comments have questions that I'm waiting for feedback on. Thanks again for this thorough review!)

NHDaly:
(One other change: I marked `coefficient` as `Base.@pure`, to ensure that it's always available at compile-time to insert into the `Val()` statement in `fldmod_by_const`.)

Without this, I was seeing type instability for `f>=3`.

NHDaly:
**Excellent** news! We can now close this PR because Julia+LLVM now do this optimization themselves, automatically! 🎉 

Since Julia 1.6, when `*(::FD{Int64}, ::FD{Int64})` took ~8ns, now it takes 0.8 ns, and the fld-by-const is optimized away. 😎 

Before:
Screenshot 2024-06-12 at 2 21 39 PM

After:
Screenshot 2024-06-12 at 2 22 00 PM

(The image is scaled to fit the browser width, but the total time is 10x less)

PR #47, opened 2019-02-04
Drop support for Julia 0.6

omus:


Project.toml
```diff
@@ -0,0 +1,13 @@
+name = "FixedPointDecimals"
+uuid = "fb4d412d-6eee-574d-9565-ede6634db7b0"
+authors = ["Fengyang Wang <fengyang.wang.0@gmail.com>", "Curtis Vogt <curtis.vogt@gmail.com>"]
+
+[compat]
+julia = "0.7, 1.0, 1.1"
```
ararslan: 1.1 is covered by 1.0, oddly enough. `julia = "0.7, 1.0"` is [0.7, 0.8) union [1.0, 2.0).

test/utils.jl
```diff
@@ -42,8 +42,4 @@ function ceil_alt(::Type{FD{T,f}}, val::AbstractFloat) where {T<:Integer, f}
     reinterpret(FD{T,f}, copysign(v + (s > 0 ? r : zero(T)), s))
 end
 
-if VERSION < v"0.7.0-DEV.4524"
-    sprintcompact(x...) = sprint(showcompact, x...)
-else
-    sprintcompact(x...) = sprint(show, x..., context=:compact=>true)
-end
+sprintcompact(x...) = sprint(show, x..., context=:compact=>true)
```
ararslan: Can we just change the uses in the tests to use `sprint(show, x..., context=:compact=>true)` directly?

Project.toml
omus: Interesting. TIL

test/utils.jl
omus: Yes, it wasn't as bad as I thought it would be.

NHDaly:
:) thanks for doing this!

I was under the impression that we're still not yet supposed to actually include `.toml` files in publicly registered packages. Has that changed?

ararslan:
> Has that changed?

Yep! All's well now. It used to confuse JuliaCIBot, which provided one of the METADATA PR checks, but that has since been fixed (and JuliaCIBot has since been taken down).

PR #51, opened 2020-02-01
Add support for three-arg `div` to FixedDecimals in Julia 1.4+

### LINKED ISSUE #50: Add support for rounding modes for division for FixedDecimals
Julia 1.4 (https://github.com/JuliaLang/julia/pull/33040) adds a standard interface to specify the rounding mode for integer division `div(x, y, ::RoundingMode)`. FixedDecimals should implement this interface as well, to support different rounding modes for division.

We've wanted to support rounding divisions for a while, anyway, so this a nice motivator.

This was originally discussed in this issue:
https://github.com/JuliaLang/julia/issues/33651

NHDaly:
This PR adds support for three-arg div by implementing `div(x,y,::RoundingMode)` for FixedDecimals that delegates to `div` of the fields.

**Remaining question:** Is this enough? Do we also need to implement the three-arg `fld`, `mod`, etc?

Fixes #50.

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -300,12 +300,14 @@ end
 for remfn in [:rem, :mod, :mod1, :min, :max]
     @eval $remfn(x::T, y::T) where {T <: FD} = reinterpret(T, $remfn(x.i, y.i))
 end
-for divfn in [:div, :fld, :fld1]
+for divfn in [:div, :fld, :fld1, :cld]
     # div(x.i, y.i) eliminates the scaling coefficient, so we call the FD constructor.
     # We don't need any widening logic, since we won't be multiplying by the coefficient.
-    @eval $divfn(x::T, y::T) where {T <: FD} = T($divfn(x.i, y.i))
+    @eval Base.$divfn(x::T, y::T) where {T <: FD} = T($divfn(x.i, y.i))
 end
 if VERSION >= v"1.4.0-"
+    # div(x.i, y.i) eliminates the scaling coefficient, so we call the FD constructor.
+    # We don't need any widening logic, since we won't be multiplying by the coefficient.
     Base.div(x::T, y::T, r::RoundingMode) where {T <: FD} = T(div(x.i, y.i, r))
 end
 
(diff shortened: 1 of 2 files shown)


src/FixedPointDecimals.jl
```diff
@@ -305,6 +305,9 @@ for divfn in [:div, :fld, :fld1]
     # We don't need any widening logic, since we won't be multiplying by the coefficient.
```
TotalVerb: Let's add `cld` to this loop for completeness, and a comment to remove it when the package is updated for Julia 2.0, since the fallbacks won't be needed then. For now these fallbacks are still necessary on older versions and on newer versions they don't do any harm.

test/runtests.jl
```diff
@@ -536,6 +536,16 @@ end
 
         @test one(FD{Int32, 2}) ÷ one(FD{Int64, 6}) isa FD{Int64, 6}
     end
+
+    @testset "div with rounding modes" begin
+        if VERSION >= v"1.4.0-"
+            @testset for x in keyvalues[FD2]
+                for R in (RoundUp, RoundDown, RoundNearest, RoundNearestTiesAway)
+                    @test div(x, 2one(x), R) === div(x, 2, R) === FD2(div(x.i, FD2(2).i, R))
```
TotalVerb: `RoundToZero` should be added here also for completeness, even though it's redundant with the non-parameterized `div`.

NHDaly: I skipped `RoundToZero` and `RoundFromZero`, since i wasn't sure if those are supposed to be supported or not?

-- No one has responded on my issue noting that `RoundFromZero` is not supported for ints:
https://github.com/JuliaLang/julia/issues/34519

TotalVerb: `RoundToZero` is the default mode, so it should be supported.

src/FixedPointDecimals.jl
NHDaly: Done. Thanks!

-----------

@TotalVerb -- Good call. `cld` was broken before this change.
Before:
```julia

julia> let FD2 = FixedDecimal{Int,2}
           cld(FD2(0.01), FD2(2))
       end
FixedDecimal{Int64,2}(0.00)
```
after:
```julia
julia> let FD2 = FixedDecimal{Int,2}
           cld(FD2(0.01), FD2(2))
       end
FixedDecimal{Int64,2}(1.00)
```

test/runtests.jl
NHDaly: Okay cool, makes sense. Done. And i've added a comment for `RoundFromZero` once the issue is addressed.

PR #52, opened 2020-02-01
Explicitly extend Base functions (Base.foo(...) = ...) instead of imports.

NHDaly:
Explicitly extend Base functions (via `Base.foo(...) = ...`) instead of importing the names to be extended.

Switches to `using Base: ...` for non-extending imports (only for `decompose`).

Style consensus seems to be that extending via `imports` should be
discouraged since it can lead to both accidentally extending functions,
and accidentally _not_ extending the intended functions.

This is a style-only change; there should be no substantive change at all.

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -500,6 +500,7 @@ function max_exp10(::Type{T}) where {T <: Integer}
 end
 
 max_exp10(::Type{BigInt}) = -1
+
 # Freeze the evaluation for BitInteger types, since max_exp10() is too compilicated to get
 # optimized away by the compiler during const-folding. (We can't freeze for user-defined
 # types because we don't know what they are yet.)
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -343,7 +343,7 @@ Base.one(::Type{FD{T, f}}) where {T, f} = reinterpret(FD{T, f}, coefficient(FD{T
 
 # comparison
 Base.:(==)(x::T, y::T) where {T <: FD} = x.i == y.i
-Base.:( <)(x::T, y::T) where {T <: FD} = x.i  < y.i
+Base.:(<)(x::T, y::T) where {T <: FD} = x.i < y.i
 Base.:(<=)(x::T, y::T) where {T <: FD} = x.i <= y.i
 
 # predicates and traits
@@ -480,10 +480,6 @@ NOTE: This function is expensive, since it contains a while-loop, but it is actu
       This function does not have or depend on any side-effects.
 """
 function max_exp10(::Type{T}) where {T <: Integer}
-    # This function is marked as `@pure`. Even though it does call some generic
-    # functions, they are all simple methods that should be able to be evaluated as
-    # constants. This function does not have or depend on any side-effects.
-
     W = widen(T)
     type_max = W(typemax(T))
 
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -27,10 +27,7 @@ module FixedPointDecimals
 
 export FixedDecimal, RoundThrows
 
-import Base: reinterpret, zero, one, abs, sign, ==, <, <=, +, -, /, *, div, rem, divrem,
-             fld, mod, fldmod, fld1, mod1, fldmod1, isinteger, typemin, typemax,
-             print, show, string, convert, parse, promote_rule, min, max,
-             floatmin, floatmax, trunc, round, floor, ceil, eps, float, widemul, decompose
+using Base: decompose
 
 const BitInteger = Union{Int8, UInt8, Int16, UInt16, Int32, UInt32, Int64,
                          UInt64, Int128, UInt128}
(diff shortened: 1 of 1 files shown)


NHDaly:
@TotalVerb I opened this as a follow-up to the last PR (#51), since i accidentally messed up at first by creating a new, local `FixedPointDecimals.cld()` function instead of extending `Base.cld` like I meant to.

src/FixedPointDecimals.jl
```diff
@@ -298,7 +295,7 @@ function convert(::Type{FD{T, f}}, x::FD{U, g}) where {T, f, U, g}
 end
 
 for remfn in [:rem, :mod, :mod1, :min, :max]
-    @eval $remfn(x::T, y::T) where {T <: FD} = reinterpret(T, $remfn(x.i, y.i))
+    @eval Base.$remfn(x::T, y::T) where {T <: FD} = reinterpret(T, $remfn(x.i, y.i))
```
TotalVerb: By the way, `rem` also supports a third `RoundingMode` argument, so we should treat it like `div` and implement a 3-argument version. Sorry for missing this.

NHDaly: No problem. I should've asked about it; i knew that at one point, but i forgot since the last time I looked at this...

NHDaly: I've opened https://github.com/JuliaMath/FixedPointDecimals.jl/pull/53 for this, since it's proven to be somewhat complicated.

Please ignore this for now in this PR! :)

src/FixedPointDecimals.jl
```diff
@@ -475,8 +469,17 @@ end
...
+      We achieve this by `@eval`ing new methods in a loop, below. Users can do this
+      themselves to add more "frozen" methods for custom Integer types:
+      ```julia
+      @eval FixedPointDecimals.max_exp10(::Type{CustomIntType}) = \$(max_exp10(CustomIntType))
+      ```
+      This function does not have or depend on any side-effects.
 """
-Base.@pure function max_exp10(::Type{T}) where {T <: Integer}
+function max_exp10(::Type{T}) where {T <: Integer}
     # This function is marked as `Base.@pure`. Even though it does call some generic
```
TotalVerb: comment has bitrotted

TotalVerb: other than this, LGTM

src/FixedPointDecimals.jl
```diff
@@ -114,22 +108,22 @@ floattype(::Type{<:FD{T}}) where {T<:Integer} = Float64
...
+Base.abs(x::FD{T, f}) where {T, f} = reinterpret(FD{T, f}, abs(x.i))
 
-+(x::FD{T, f}, y::FD{T, f}) where {T, f} = reinterpret(FD{T, f}, x.i+y.i)
--(x::FD{T, f}, y::FD{T, f}) where {T, f} = reinterpret(FD{T, f}, x.i-y.i)
+Base.:+(x::FD{T, f}, y::FD{T, f}) where {T, f} = reinterpret(FD{T, f}, x.i+y.i)
+Base.:-(x::FD{T, f}, y::FD{T, f}) where {T, f} = reinterpret(FD{T, f}, x.i-y.i)
 
 # wide multiplication
-Base.@pure function widemul(x::FD{<:Any, f}, y::FD{<:Any, g}) where {f, g}
+Base.@pure function Base.widemul(x::FD{<:Any, f}, y::FD{<:Any, g}) where {f, g}
```
omus: I would just do `using Base: @pure` to clean things up

src/FixedPointDecimals.jl
```diff
@@ -348,24 +342,24 @@ Base.zero(::Type{FD{T, f}}) where {T, f} = reinterpret(FD{T, f}, zero(T))
 Base.one(::Type{FD{T, f}}) where {T, f} = reinterpret(FD{T, f}, coefficient(FD{T, f}))
 
 # comparison
-==(x::T, y::T) where {T <: FD} = x.i == y.i
- <(x::T, y::T) where {T <: FD} = x.i  < y.i
-<=(x::T, y::T) where {T <: FD} = x.i <= y.i
+Base.:(==)(x::T, y::T) where {T <: FD} = x.i == y.i
+Base.:( <)(x::T, y::T) where {T <: FD} = x.i  < y.i
```
omus: I say we just drop the whitespace padding here
```suggestion
Base.:(<)(x::T, y::T) where {T <: FD} = x.i < y.i
```

src/FixedPointDecimals.jl
```diff
@@ -497,9 +500,12 @@ Base.@pure function max_exp10(::Type{T}) where {T <: Integer}
 end
 
 max_exp10(::Type{BigInt}) = -1
-# Freeze the evaluation for Int128, since max_exp10(Int128) is too compilicated to get
-# optimized away by the compiler during const-folding.
-@eval max_exp10(::Type{Int128}) = $(max_exp10(Int128))
+# Freeze the evaluation for BitInteger types, since max_exp10() is too compilicated to get
```
omus:
```suggestion

# Freeze the evaluation for BitInteger types, since max_exp10() is too compilicated to get
```

src/FixedPointDecimals.jl
NHDaly: Done. Thanks

src/FixedPointDecimals.jl
NHDaly: Oops. I don't think i actually meant for this change to be part of this PR... BUT i guess it doesn't hurt. It is definitely an improvement, so i'll just carry it along. Sorry about that; thanks for the careful review.

NHDaly: Fixed comment. Done.

PR #53, opened 2020-02-01
Add support for rem(x,y,::RoundingMode) for FixedDecimals

NHDaly:
However, the simple overload does not seem to work, since it is
ambiguous with `rem(x,y,::RoundingMode{Up})`, for example, so for `rem`,
we're manually overloading all RoundingModes.

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -294,6 +294,15 @@ end
 for remfn in [:rem, :mod, :mod1, :min, :max]
     @eval Base.$remfn(x::T, y::T) where {T <: FD} = reinterpret(T, $remfn(x.i, y.i))
 end
+if VERSION >= v"1.4.0-"
+    # TODO: Add support for (RoundFromZero, RoundNearest, RoundNearestTiesAway), which
+    #       aren't supported for Ints.
+    for R in (RoundToZero, RoundUp, RoundDown)
+        RType = typeof(R)
+        Base.rem(x::T, y::T, r::RType) where {T <: FD} = reinterpret(T, rem(x.i, y.i, r))
+    end
+end
+
 # TODO: When we upgrade to a min julia version >=1.4 (i.e Julia 2.0), this block can be
 # dropped in favor of three-argument `div`, below.
 for divfn in [:div, :fld, :fld1, :cld]
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -537,13 +537,26 @@ end
         @test one(FD{Int32, 2}) ÷ one(FD{Int64, 6}) isa FD{Int64, 6}
     end
 
+    @testset "rem with rounding modes" begin
+        if VERSION >= v"1.4.0-"
+            # TODO: Test RoundFromZero, RoundNearest, RoundNearestTiesAway
+            #       See: https://github.com/JuliaLang/julia/issues/34519
+            @testset for x in keyvalues[FD2],
+                         R in (RoundToZero, RoundUp, RoundDown)
+                @test rem(x, 2one(x), R) === rem(x, 2, R) === reinterpret(FD2, rem(x.i, FD2(2).i, R))
+            end
+        end
+        @testset for x in keyvalues[FD2], f in (fld, cld, fld1, div)
+            @test f(x, 2one(x)) === f(x, 2) === FD2(f(x.i, FD2(2).i))
+        end
+    end
+
     @testset "div with rounding modes" begin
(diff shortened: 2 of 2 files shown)


NHDaly:
@TotalVerb I originally added the straightforward method, below, but it gave a MethodAmbiguity error:
```julia
julia> Base.rem(x::T, y::T, r::RoundingMode) where {T <: FD} = reinterpret(T, rem(x.i, y.i, r))
```
```julia
julia> rem(x, x, RoundUp)
ERROR: MethodError: rem(::FixedDecimal{Int64,2}, ::FixedDecimal{Int64,2}, ::RoundingMode{:Up}) is ambiguous. Candidates:
  rem(x::T, y::T, r::RoundingMode) where T<:FixedDecimal in Main at REPL[20]:1
  rem(x, y, ::RoundingMode{:Up}) in Base at div.jl:69
Possible fix, define
  rem(::T, ::T, ::RoundingMode{:Up}) where T<:FixedDecimal
```

So i implemented all the more specific methods instead.

This actually leads me to wonder whether we should be doing the `div()` methods the same way...

-----------------------------

And in fact, the generic definition of `rem` provided in Base, `rem(x, y, ::RoundingMode{:Up}) = mod(x,-y)` means that we actually already support the 3-arg `rem` out of the box without doing anything more in this package... I now remember that i knew this already, which is why i didn't implement it last time -- i wanted to ask someone about this.

Those defs are here:
https://github.com/JuliaLang/julia/blob/fbc2c0aec19a7c415011c1abd4f05622ac9c41e7/base/div.jl#L67-L69


I'm wondering if we're not supposed to be overloading three-argument `div` and `rem` for generic `::RoundingMode`s? The API isn't super clear or well-documented.

NHDaly:
Additionally, one other concern:

The default 3-arg `Base.rem` implementation provided above doesn't call `promote` like the 2-arg versions do, which means that our newly added methods aren't being triggered for non-FixedDecimal types. I could probably add the overloads to call promote, but this might be worth filing with julia itself?
```julia
julia> @which rem(FD2(0.5), FD2(2), RoundToZero)
rem(x::T, y::T, r::RoundingMode{:ToZero}) where T<:FixedDecimal in FixedPointDecimals at /Users/daly/julia-dev/FixedPointDecimals/src/FixedPointDecimals.jl:302

julia> @which rem(FD2(0.5), 2, RoundToZero)
rem(x, y, ::RoundingMode{:ToZero}) in Base at div.jl:67
```

TotalVerb:
This seems to be an unfortunate inconsistency within Julia itself (where the `div` 3-arg is being made primitive, with `fld`, etc. deferring to it, while the `rem` 3-arg is currently non-primitive and calls `rem`/`mod`). If the `# TODO` resolves in Base, then we will need to implement something like this PR, but until then I suppose it is not needed.

src/FixedPointDecimals.jl
```diff
@@ -294,6 +294,15 @@ end
 for remfn in [:rem, :mod, :mod1, :min, :max]
     @eval Base.$remfn(x::T, y::T) where {T <: FD} = reinterpret(T, $remfn(x.i, y.i))
 end
+if VERSION >= v"1.4.0-"
+    # TODO: Add support for (RoundFromZero, RoundNearest, RoundNearestTiesAway), which
+    #       aren't supported for Ints.
+    for R in (RoundToZero, RoundUp, RoundDown)
+        RType = typeof(R)
+        Base.rem(x::T, y::T, r::RType) where {T <: FD} = reinterpret(T, rem(x.i, y.i, r))
```
omus: Why not just do?
```suggestion
        Base.rem(x::T, y::T, r::Type{R}) where {T <: FD} = reinterpret(T, rem(x.i, y.i, r))
```

TotalVerb: it would have to be `typeof(R)`, not `Type{R}` here (rounding modes are singleton values, not types)

omus: You are correct. I've updated my suggestion appropriately

NHDaly: 👍 thanks @omus. yes agreed this is much simpler. i'm not sure why i did it the other way at first. Done.

NHDaly:
> If the `# TODO` resolves in Base, then we will need to implement something like this PR, but until then I suppose it is not needed.

Okay yeah, makes sense. Then i'm going to put this PR on hold until such time as it's resolved in Base?

TotalVerb:
That sounds reasonable to me. Thanks for your hard work!

Issue #56, opened 2020-06-09
Add support for `numerator(::FD)` and `denominator(::FD)`


NHDaly:
We already support converting a FixedDecimal to a Rational, via:
https://github.com/JuliaMath/FixedPointDecimals.jl/blob/e8c9f3b4b68c85fa0bb8813b35c27ee267114175/src/FixedPointDecimals.jl#L324-L326

We should also provide `numerator()` and `denominator()`, as part of that interface. They might be implemented like this:
```julia
function Base.numerator(x::FixedDecimal{T,f})::T where {T, f}
    return value(x)
end
function Base.denominator(x::FixedDecimal{T,f})::T where {T, f}
    return coefficient(x)
end
```

PR #57, opened 2020-06-28
add func & test for getting numerator & denominator of fixed point decimal number

michiboo:
Address issues #56


NHDaly:
This seems nice to me. Thanks for the PR! Many apologies for our very long delay.

It's probably worth considering whether there could be a more performant way to do this..

For example, we could start by constructing the rational via `Rational(x.i, coefficient(FD{T, f}))`, and _then_ take its num and dec. I think that would probably be more performant, since it avoids unnecessarily converting to a Float in between?
```julia
    return Rational(x.i, coefficient(FD{T, f})).num
```
Something like that, i think, anyway.

NHDaly (approved):
That said, i'm happy to merge like this if we want to, as well

Issue #59, opened 2020-10-20
Project.toml missing a `version` entry


Sacha0:
On Julia 1.6-dev (57e6bead349117417c3b0b0ad4fe1b248197fee5, tip on 10/20), on `up` Pkg complains
```
ERROR: project file for FixedPointDecimals is missing a `version` entry
```
Best! :)

Sacha0:
Fixed by #60! :)

PR #60, opened 2020-10-20
Add version to Project.toml, and bump to 0.3.1.

Sacha0:
This pull request adds `version` to Project.toml, and bumps `version` to 0.3.1 for downstream tagging/registration. These changes should address https://github.com/JuliaMath/FixedPointDecimals.jl/issues/59, i.e. functionality under Julia 1.6. Best! :)


Sacha0:
@NHDaly ^ :)

NHDaly:
wow, crazy pants that we didn't have this. thanks @Sacha0! :)

Sacha0:
@rai-nhdaly, I just noticed that I think we still need to tag and register with this patch in? I lack the power, hence the ping :).

PR #69, opened 2022-02-16
removed @pure's

caseykneale:
Working with @NHDaly we found that the pure annotations didn't change lowered code in most circumstances(Julia 1.6.5), likely due to improvements in the Julia compiler. The @pure annotations however do introduce potential trouble when used outside of their appropriate contexts (https://github.com/JuliaLang/julia/pull/39954#issuecomment-812876814).

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -27,7 +27,7 @@ module FixedPointDecimals
 
 export FixedDecimal, RoundThrows
 
-using Base: decompose, BitInteger, @pure
+using Base: decompose, BitInteger
 
 # floats that support fma and are roughly IEEE-like
 const FMAFloat = Union{Float16, Float32, Float64, BigFloat}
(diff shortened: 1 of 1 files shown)


omus:
I'm good for this as long as @NHDaly approves. It may be reasonable to also update the minimum version of Julia required here if we think this could impact older version of Julia negatively. We should probably drop support for Julia versions before Julia 1.6 anyway so that works out.

NHDaly:
Yeah, i'm good for this. Frankly, i just didn't know what I was doing back when I introduced all those `@pure`s. I think that they're all probably correct/fine as long as you only used FixedDecimal with types from base, like `Int64` but they're definitely not fine if you use it with your own custom Integer type.


I had been meaning to go back and delete all these `@pure` annotations for a long time, once I understood that they're a safety violation. I'll feel much better about us finally tagging a release once these are gone! 😊

NHDaly (approved):
Even if this turns out to be a performance regression, I believe it's worth it because we were previously violating safety rules. 👍 

We can reinvestigate how to improve the performance if we ever find ourselves needing to come back to this!

NHDaly:
Oh shoot, sorry @omus: I forgot about your comment to first raise the lower-bound on the Julia version. Lemme do that now, too, in a separate PR.

NHDaly:
🙄 and then i accidentally committed directly to master. Anyway, i've done that here:
https://github.com/JuliaMath/FixedPointDecimals.jl/commit/460eb8ff46c2a01ccb6c5363c75c7e5321c33669

I think we could make a release now, too! I'll go ahead and do that 🎉

PR #78, opened 2023-03-08
Implement `round` methods for all `RoundingMode`s

bachdavi:
This PR adds support for rounding with:

- `RoundUp` → alias for `ceil`
- `RoundDown` → alias for `floor`
- `RoundToZero` → alias for `trunc`

`RoundNearestTiesAway` and `RoundNearestTiesUp` are implemented in `_round_to_even`.

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -125,16 +125,19 @@ end
 Base.widemul(x::Integer, y::FD) = widemul(y, x)
 
 """
-    _round_to_even(quotient, remainder, divisor, ::RoundingMode{M})
+    _round_to_nearest(quotient, remainder, divisor, ::RoundingMode{M})
 
-Round `quotient + remainder / divisor` to the nearest even integer,
+Round `quotient + remainder / divisor` to the nearest integer,
 given that `0 ≤ remainder < divisor` or `0 ≥ remainder >
 divisor`. (This assumption is satisfied by the return value of
 `fldmod` in all cases, and the return value of `divrem` in cases where
 `divisor` is known to be positive.) The tie is decided depending on
 the `RoundingMode`.
 """
-function _round_to_even(quotient::T, remainder::T, divisor::T, ::RoundingMode{M}=RoundNearest) where {T <: Integer, M}
+function _round_to_nearest(quotient::T,
+                           remainder::T,
+                           divisor::T,
+                           ::RoundingMode{M}=RoundNearest) where {T <: Integer, M}
     halfdivisor = divisor >> 1
     if iseven(divisor) && remainder == halfdivisor
         # `:NearestTiesAway` will tie away from zero, e.g. -8.5 ->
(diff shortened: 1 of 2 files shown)


src/FixedPointDecimals.jl
```diff
@@ -125,24 +125,36 @@ end
...
+        # -9. `:NearestTiesUp` will always ties towards positive
+        # infinity. `:Nearest` will tie towards the nearest even
+        # integer.
+        if M == :NearestTiesAway && quotient < zero(quotient)
+            quotient
+        elseif M == :Nearest && iseven(quotient)
+            quotient
+        else
+            quotient + one(quotient)
+        end
```
NHDaly, line 150: I _think_ the `ifelse()` here was probably a perf optimization. We should be careful about changing it.

From my understanding the tradeoff is this:
With `ifelse()`, you always compute both possible values (so you do some "extra" computation) in exchange for avoiding the branch. Since branches are more expensive than the one addition, it's worth it to do the `ifelse`.

((I'm not sure; maybe LLVM would do this anyway though.))

---------

That said, the check on `M` can compile away, so it's just the second parts of those conditions that we need in the ifelse..

I think we can retain the structure of the old code like this:
```suggestion
        if M == :NearestTiesAway
            ifelse(quotient < zero(quotient), quotient, quotient + one(quotient))
        elseif M == :Nearest
            ifelse(iseven(quotient), quotient, quotient + one(quotient))
        else
            quotient + one(quotient)
        end
```

src/FixedPointDecimals.jl
```diff
@@ -125,24 +125,36 @@ end
...
+
+Round `quotient + remainder / divisor` to the nearest even integer,
+given that `0 ≤ remainder < divisor` or `0 ≥ remainder >
+divisor`. (This assumption is satisfied by the return value of
+`fldmod` in all cases, and the return value of `divrem` in cases where
+`divisor` is known to be positive.) The tie is decided depending on
+the `RoundingMode`.
 """
-function _round_to_even(quotient::T, remainder::T, divisor::T) where {T <: Integer}
+function _round_to_even(quotient::T, remainder::T, divisor::T, ::RoundingMode{M}=RoundNearest) where {T <: Integer, M}
```
NHDaly, line 137: 🤔 I guess the name of the function should change, since it's not only doing "round to nearest even integer" anymore, right? "round to even" is what `RoundNearest` does:

```
help?> RoundNearest
search: RoundNearest RoundNearestTiesUp RoundNearestTiesAway

  RoundNearest

  The default rounding mode. Rounds to the nearest integer, with ties
  (fractional values of 0.5) being rounded to the nearest even integer.
```

So the docstring could change as well. It's now just "round to the nearest integer" (not nearest _even_ integer), with the ties decided by the rounding mode.

(In fact, the old name was confusing, and probably should have been `_round_ties_to_even`)

NHDaly (approved):
LGTM, thanks for the PR, @bachdavi!

src/FixedPointDecimals.jl
bachdavi: 👍 
@NHDaly Thank you for the review!

bachdavi:
@NHDaly When you find time can you hit the merge button :) 
I don't have write access, it would be best to do a release and then I'll PR the changes on our side.
I'll add an issue on Jira.

NHDaly:
Oops, thanks for the reminder! I got the open source vs internal protocols mixed up. thanks! Merged and created a release

PR #80, opened 2023-06-13
Use a `Parsers.jl`-based parser implementation

Drvi:
This hooks into the floating point parsing machinery from `Parsers.jl`, where we also accumulate all the digits and note the effective exponent before we do "scaling" -- for `FixedDecimals`, the scaling means padding the backing integer with zeros or rounding them as necessary. This change should improve performance noticeably as we no longer allocate any strings during parsing.

`RoundNearest` rounding was redone during parsing to match the rounding behaviors on already parsed integers.

cc: @quinnj @NHDaly

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,6 +7,7 @@ version = "0.4.2"
 Parsers = "69de0a69-1ddd-5017-9359-2bf0b02dc9f0"
 
 [compat]
+Parsers = "2.7"
 julia = "1.6"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,9 @@ uuid = "fb4d412d-6eee-574d-9565-ede6634db7b0"
 authors = ["Fengyang Wang <fengyang.wang.0@gmail.com>", "Curtis Vogt <curtis.vogt@gmail.com>"]
 version = "0.4.2"
 
+[deps]
+Parsers = "69de0a69-1ddd-5017-9359-2bf0b02dc9f0"
+
 [compat]
 julia = "1.6"
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "FixedPointDecimals"
 uuid = "fb4d412d-6eee-574d-9565-ede6634db7b0"
 authors = ["Fengyang Wang <fengyang.wang.0@gmail.com>", "Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "0.4.2"
+version = "0.5.0"
 
 [deps]
 Parsers = "69de0a69-1ddd-5017-9359-2bf0b02dc9f0"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "FixedPointDecimals"
 uuid = "fb4d412d-6eee-574d-9565-ede6634db7b0"
 authors = ["Fengyang Wang <fengyang.wang.0@gmail.com>", "Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "0.5.0"
+version = "0.4.3"
 
 [deps]
 Parsers = "69de0a69-1ddd-5017-9359-2bf0b02dc9f0"
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -100,12 +100,15 @@ end
 end
 
 const FD = FixedDecimal
+const RoundThrows = RoundingMode{:Throw}()
 
 include("parse.jl")
 
 function __init__()
-    append!(_BIGINT_10s, [BigInt(0) for i in 0:Threads.nthreads()])
-    append!(_BIGINT_Rs, [BigInt(0) for i in 0:Threads.nthreads()])
+    nt = isdefined(Base.Threads, :maxthreadid) ? Threads.maxthreadid() : Threads.nthreads()
+    # Buffers used in parsing when dealing with BigInts, see _divpow10! in parse.jl
+    resize!(empty!(_BIGINT_10s), nt)
+    resize!(empty!(_BIGINT_Rs), nt)
 end
 
 (::Type{T})(x::Real) where {T <: FD} = convert(T, x)
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -100,7 +100,6 @@ end
 end
 
 const FD = FixedDecimal
-const RoundThrows = RoundingMode{:Throw}()
 
 include("parse.jl")
 
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -28,6 +28,7 @@ module FixedPointDecimals
 export FixedDecimal, RoundThrows
 
 using Base: decompose, BitInteger
+import Parsers
 
 # floats that support fma and are roughly IEEE-like
 const FMAFloat = Union{Float16, Float32, Float64, BigFloat}
@@ -99,6 +100,13 @@ end
 end
 
 const FD = FixedDecimal
+
+include("parse.jl")
+
+function __init__()
+    append!(_BIGINT_10s, [BigInt(0) for i in 0:Threads.nthreads()])
+    append!(_BIGINT_Rs, [BigInt(0) for i in 0:Threads.nthreads()])
+end
 
 (::Type{T})(x::Real) where {T <: FD} = convert(T, x)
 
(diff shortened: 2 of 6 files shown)


src/FixedPointDecimals.jl
```diff
@@ -413,78 +425,6 @@ function Base.show(io::IO, x::FD{T, f}) where {T, f}
...
 end
 
-# parsing
-
-"""
-    RoundThrows
-
-Raises an `InexactError` if any rounding is necessary.
-"""
-const RoundThrows = RoundingMode{:Throw}()
```
NHDaly, line 423: Put back the docstring

src/FixedPointDecimals.jl
```diff
@@ -99,6 +100,17 @@ end
 end
 
 const FD = FixedDecimal
+const RoundThrows = RoundingMode{:Throw}()
```
NHDaly, line 103: oh actually, it's just defined twice. You can delete one or the other :)

src/parse.jl
```diff
@@ -0,0 +1,227 @@
...
+        # Check if the input is empty
+        if Parsers.eof(source, pos, len)
+            code |= Parsers.INVALID | Parsers.EOF
+            x = zero(T)
+            @goto done
+        end
+    end
+
+    if (b - UInt8('0')) <= 0x09 || b == options.decimal
+        x, code, pos = Parsers.parsedigits(conf, source, pos, len, b, code, options, UInt64(0), neg, startpos, true, 0, nothing)
```
NHDaly, line 189: This definitely needs some comments explaining the flow.
something like:
```
        # Share the floating point parser from Parsers, returns the backing integer
```
explaining:
- how the flow goes from here
    - we call into parsers, wihch then calls back into our functions?
    - which overloads this calls i guess
- what is returned

src/parse.jl
```diff
@@ -0,0 +1,227 @@
...
+    # Positive: how many trailing zeroes we need to add to out integer
+    # Negative: how many digits are past our precision (we need to handle them in rounding)
+    decimal_shift = conf.f + exp
+    # Number of digits we need to accumulate including any trailigng zeros or digits past our precision
+    backing_integer_digits = ndigits + decimal_shift
+    may_overflow = backing_integer_digits == _maxintdigits(T)
+    if iszero(ndigits)
+        # all digits are zero
+        i = zero(T)
+    elseif backing_integer_digits < 0
```
NHDaly, line 114:
```suggestion
    # The backing_integer_digits == 0 case is handled in the `else` (it means
    # that all the digits are passed the precision but we might get `1` from rounding) 
    elseif backing_integer_digits < 0
```

src/parse.jl
```diff
@@ -0,0 +1,227 @@
...
+    y == 0 || (code |= Parsers.INEXACT)
+    return x, code
+end
+
+# Rescale the digits we accumulated so far into the the a an integer representing the decimal
+@inline function Parsers.scale(
+    conf::FixedDecimalConf{T}, ::Parsers.FloatType, digits::V, exp, neg, code, ndigits, f::F, options::Parsers.Options
+) where {T,V,F}
+    rounding = something(options.rounding, RoundThrows)
+    # Positive: how many trailing zeroes we need to add to out integer
```
NHDaly, line 105:
```suggestion
    # Positive: how many trailing zeroes we need to add to our integer
```

src/parse.jl
```diff
@@ -0,0 +1,227 @@
...
+    Base.GMP.MPZ.pow_ui!(y, pow)      # y = y^pow
+    Base.GMP.MPZ.tdiv_qr!(x, y, x, y) # x, y = divrem(x, y)
+    y == 0 || (code |= Parsers.INEXACT)
+    return x, code
+end
+
+# Rescale the digits we accumulated so far into the the a an integer representing the decimal
+@inline function Parsers.scale(
+    conf::FixedDecimalConf{T}, ::Parsers.FloatType, digits::V, exp, neg, code, ndigits, f::F, options::Parsers.Options
+) where {T,V,F}
```
NHDaly, line 103: Needs a comment: Why is `FloatType` here?

src/FixedPointDecimals.jl
quinnj:
```suggestion
```

src/parse.jl
```diff
@@ -0,0 +1,227 @@
+using Parsers
+using Parsers: AbstractConf, SourceType, XOPTIONS, Result
```
quinnj, line 2:
```suggestion
using Parsers: AbstractConf, Result
```

src/parse.jl
```diff
@@ -0,0 +1,227 @@
...
+
+"""
+    RoundThrows
+
+Raises an `InexactError` if any rounding is necessary.
+"""
+const RoundThrows = RoundingMode{:Throw}()
+
+struct FixedDecimalConf{T<:Integer} <: AbstractConf{T}
```
quinnj, line 11:
```suggestion
# make our own conf struct to avoid specializing Parsers.typeparser on each unique precision value
struct FixedDecimalConf{T<:Integer} <: AbstractConf{T}
```

src/parse.jl
```diff
@@ -0,0 +1,227 @@
...
+    f::Int
+end
+Parsers.conf(::Type{FixedDecimal{T,f}}, opts::Parsers.Options, kw...) where {T<:Integer,f} = FixedDecimalConf{T}(f)
```
quinnj, line 14:
```suggestion
# this overload says that when parsing a FixedDecimal type, use our new custom FixedDecimalConf type
Parsers.conf(::Type{FixedDecimal{T,f}}, opts::Parsers.Options, kw...) where {T<:Integer,f} = FixedDecimalConf{T}(f)
```

src/parse.jl
```diff
@@ -0,0 +1,227 @@
...
+Parsers.returntype(::Type{FixedDecimal{T,f}}) where {T,f} = T
```
quinnj, line 15:
```suggestion
# Because the value returned from our `typeparser` isn't a FixedDecimal, we overload here to show we're returning an integer type
Parsers.returntype(::Type{FixedDecimal{T,f}}) where {T,f} = T
```

src/parse.jl
```diff
@@ -0,0 +1,227 @@
...
+function _divpow10!(x::BigInt, code, pow, ::RoundingMode{:Throw})
+    @inbounds y = Parsers.access_threaded(() -> (@static VERSION > v"1.5" ? BigInt(; nbits=256) : BigInt()), _BIGINT_10s) # we must not yield here!
+    Base.GMP.MPZ.set!(y, _BIGINT10)   # y = 10
+    Base.GMP.MPZ.pow_ui!(y, pow)      # y = y^pow
+    Base.GMP.MPZ.tdiv_qr!(x, y, x, y) # x, y = divrem(x, y)
+    y == 0 || (code |= Parsers.INEXACT)
+    return x, code
+end
+
+# Rescale the digits we accumulated so far into the the a an integer representing the decimal
```
quinnj, line 100:
```suggestion
# Rescale the digits we accumulated so far into the the a an integer representing the decimal
# Note the 2nd argument `FloatType` is used by Parsers.jl for _float_ parsing, but we can ignore in the fixed decimal case
```

src/parse.jl
quinnj: I suggested a comment above

src/parse.jl
```diff
@@ -0,0 +1,227 @@
...
+    # Negative: how many digits are past our precision (we need to handle them in rounding)
+    decimal_shift = conf.f + exp
+    # Number of digits we need to accumulate including any trailigng zeros or digits past our precision
+    backing_integer_digits = ndigits + decimal_shift
+    may_overflow = backing_integer_digits == _maxintdigits(T)
+    if iszero(ndigits)
+        # all digits are zero
+        i = zero(T)
+    elseif backing_integer_digits < 0
+        # All digits are past our precision, no overflow possible
```
quinnj, line 115: Maybe I'm not understanding the comment quite right, but this makes it sound like it's backwards; I was expecting a phrase like "if any digits are past our precision, overflow is possible" or the inverse of "No digits are past our precision, so no overflow possible"

src/parse.jl
```diff
@@ -0,0 +1,227 @@
...
+            i = _shift(r, decimal_shift)
+            may_overflow && (r >= i) && (code |= Parsers.OVERFLOW)
+        elseif decimal_shift < 0
+            if rounding === RoundNearest
+                r, code = _divpow10!(digits, code, -decimal_shift, RoundNearest)
+            elseif rounding === RoundToZero
+                r, code = _divpow10!(digits, code, -decimal_shift, RoundToZero)
+            else
+                r, code = _divpow10!(digits, code, -decimal_shift, RoundThrows)
+            end
```
quinnj, line 139: Shouldn't it be just as type-stable/performant to just do:
```suggestion
            r, code = _divpow10!(digits, code, -decimal_shift, rounding)
```
here?

Drvi: No, that would cause dynamic dispatch

src/parse.jl
Drvi: Overflow is not possible here but inexact error is possible if we're not okay with rounding. In my mind overflow is when we try to parse "130" `FD{Int8,0}` and inexact would happen if we tried to parse "100.1" as `FD{Int8,0}`

src/parse.jl
```diff
@@ -0,0 +1,227 @@
...
+@inline function Parsers.noscale(conf::FixedDecimalConf{T}, digits::Integer, neg::Bool, code, ndigits, f::F, options::Parsers.Options) where {T,F}
+    FT = Parsers.FLOAT64 # not used by FixedDecimal parser
+    exp = 0
+    return Parsers.scale(conf, FT, digits, exp, neg, code, ndigits, f, options)
+end
+
+# We return a value of T -- i.e. the _integer_ backing the FixedDecimal, the reintrpret needs to happen later
+@inline function Parsers.typeparser(conf::FixedDecimalConf{T}, source, pos, len, b, code, pl, options) where {T<:Integer}
+    if !(options.rounding in (nothing, RoundNearest, RoundToZero, RoundThrows))
+        throw(ArgumentError("Unhandled rounding mode $options.rounding"))
```
nickrobinson251, line 177:
```suggestion
        throw(ArgumentError("Unhandled rounding mode $(options.rounding)"))
```

Project.toml
```diff
@@ -1,9 +1,13 @@
 name = "FixedPointDecimals"
 uuid = "fb4d412d-6eee-574d-9565-ede6634db7b0"
 authors = ["Fengyang Wang <fengyang.wang.0@gmail.com>", "Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "0.4.2"
+version = "0.5.0"
```
Drvi, line 4:
```suggestion
version = "0.4.3"
```

src/parse.jl
```diff
@@ -0,0 +1,239 @@
...
+    return sizeof(T) <= sizeof(V) && (neg ? v > _unsafe_convert_int(V, typemax(T)) + one(V) : v > typemax(T))
+end
+_check_overflows(::Type{T}, v::T, neg::Bool) where {T <: Integer} = false
+
+# `x = div(x, 10^pow, mode)`; may set code |= INEXACT for RoundThrows
+# x is non-negative, pow is >= 1
+# `!` to signal we mutate bigints in-place
+function _divpow10!(x::T, code, pow, mode::RoundingMode) where {T}
+    return div(x, _shift(one(T), pow), mode), code
+end
```
NHDaly, line 69: Not something for right now, but just sharing this because it's _so super freaking cool_:
https://github.com/JuliaMath/FixedPointDecimals.jl/pull/45

Todd came up with this approach to improve the performance of dividing by power of ten. Basically the idea is that you can skip the `div`, which is a very expensive operation, by instead multiplying by (2^64 / 10 / 2^64), and since you can precompute `2^64 / 10` as a constant, and then `/ 2^64` can be done by just a bit-shift, you can skip the divide. Really clever!

But I never managed to merge that PR, so............................. :)

I think we just leave that optimization aside here too, but it's fun and something cool to think about for the future! :)

NHDaly (commented):
I'm sorry that I wasn't able to review today. I was pulled into an oncall investigation of a customer incident today, and it unexpectedly took up a bunch of my day. :[

I'm going to merge though, based on Jacob's review, and based on the preliminary review we did together over zoom! :) It does look like a much better approach.

Thanks all!

PR #84, opened 2023-11-17
Eliminate BigInt allocs for 128-bit FixedDecimal conversions.

NHDaly:
Before this PR, converting an Int to a FD{Int128} had to allocate a BigInt, which is fairly expensive:
```julia
julia> @time convert(FixedDecimal{Int128, 2}, 1)
  0.000003 seconds (7 allocations: 128 bytes)
FixedDecimal{Int128,2}(1.00)
```
After, the allocs are gone, and this is only involves the checks for overflow and the coefficient multiplication:
```julia
julia> @time convert(FixedDecimal{Int128, 2}, 1)
  0.000000 seconds
FixedDecimal{Int128,2}(1.00)
```

I've added tests to cover all corner cases and to ensure the allocations remain eliminated.

Also, while I was here, I have changed the InexactErrors you get for illegal conversions to be clearer, by reporting on the types involved in the conversion directly, rather than the internal errors you got previously:
```julia
Before: InexactError(:Int128, Int128, 34028236692093846346337460743176821145500)
 After: InexactError(:convert, FixedDecimal{Int128, 2}, 0xffffffffffffffffffffffffffffffff)

Before: InexactError: Int128(17014118346046923173168730371588410572700)
 After: InexactError: convert(FixedDecimal{Int128, 2}, 170141183460469231731687303715884105727)
```

Thanks!

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -296,7 +296,17 @@ end
 Base.convert(::Type{FD{T, f}}, x::FD{T, f}) where {T, f} = x  # Converting an FD to itself is a no-op
 
 function Base.convert(::Type{FD{T, f}}, x::Integer) where {T, f}
-    reinterpret(FD{T, f}, T(widemul(x, coefficient(FD{T, f}))))
+    C = coefficient(FD{T, f})
+    throw_inexact() = throw(InexactError(:convert, FD{T, f}, x))
+    typemin(T) <= x <= typemax(T) || throw_inexact()
+    xT = convert(T, x)  # This won't throw, since we already checked, above.
+    # Perform x * C, and check for overflow. This is cheaper than a widemul, especially for
+    # 128-bit T, since those widen into a BigInt.
+    v, overflow = Base.mul_with_overflow(xT, C)
+    if overflow
+        throw_inexact()
+    end
+    reinterpret(FD{T, f}, v)
 end
 
 Base.convert(::Type{T}, x::AbstractFloat) where {T <: FD} = round(T, x)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -314,10 +314,32 @@ end
     end
 end
 @testset "128-bit conversion performance" begin
-    #@test Base.convert(WFD2, 1) == WFD2(1)
-    #@test @allocated(Base.convert(WFD2, 1)) == 0
-end
+    # Baseline cases
+    @test @allocated(Base.convert(FD2, Int8(-1))) == 0
+    @test @allocated(Base.convert(FD2, UInt8(1))) == 0
+    @test @allocated(Base.convert(SFD2, Int8(-1))) == 0
 
+    # Int 128 cases
+    @test @allocated(Base.convert(WFD2, 1)) == 0
+    @test @allocated(Base.convert(WFD2, 1)) == 0
+    @test @allocated(Base.convert(UWFD2, 1)) == 0
+    @test @allocated(Base.convert(WFD2, -1)) == 0
+    # @test @allocated(Base.convert(UWFD2, -1)) == 32   # not sure how best to test this
+    @test @allocated(Base.convert(WFD2, UInt(1))) == 0
+    @test @allocated(Base.convert(UWFD2, UInt(1))) == 0
+
+    @test @allocated(Base.convert(WFD2, Int128(1))) == 0
+    @test @allocated(Base.convert(UWFD2, Int128(1))) == 0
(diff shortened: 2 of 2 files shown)


src/FixedPointDecimals.jl
```diff
@@ -296,7 +296,35 @@ end
...
+    typemin(T) <= x <= typemax(T) || throw_inexact()
+    xT = convert(T, x)  # This won't throw, since we already checked, above.
+    # Perform x * C, and check for overflow. This is cheaper than a widemul, especially for
+    # 128-bit T, since those widen into a BigInt.
+    v = _mul_checked_overflow(throw_inexact, xT, C)
+    reinterpret(FD{T, f}, v)
+end
+function Base.convert(::Type{FD{BigInt, f}}, x::Integer) where {f}
+    # We specialize on f==1, since julia can't eliminate BigInt multiplication.
+    if f == 1
```
Drvi, line 310: Sorry, previously I said we can specialize the case when `f == 1` but I meant when the `coefficient` is one, so when `f == 0`

Drvi (commented):
I'll have a look at the tests on Monday

src/FixedPointDecimals.jl
NHDaly: Oh, heh, whoops! I meant the same thing (as is evident by using the BigInt value unchanged). Duh, of course it should have been f=0. Sorry I didn't catch that myself! Also, I guess I didn't add test coverage for this case 😓 I added in the BigInt handling at the end and I wasn't being as thorough then, I think. :(

NHDaly: Thanks Tomas. Fixed this in the last commit.

Drvi: Thanks! Also, the comment needs updating.

test/FixedDecimal.jl
```diff
@@ -0,0 +1,1098 @@
...
+    @test @allocated(convert(F64D2, Int8(-1))) == 0
+    @test @allocated(convert(F64D2, UInt8(1))) == 0
+    @test @allocated(convert(SFD2, Int8(-1))) == 0
+
+    # Int 128 cases
+    @test @allocated(convert(WFD2, 1)) == 0
+    @test @allocated(convert(WFD2, 1)) == 0
+    @test @allocated(convert(UWFD2, 1)) == 0
+    @test @allocated(convert(WFD2, -1)) == 0
+    # @test @allocated(convert(UWFD2, -1)) == 32   # not sure how best to test this
```
Drvi, line 363: This should help https://github.com/JuliaLang/AllocCheck.jl

Drvi (approved):
LGTM!

test/FixedDecimal.jl
NHDaly: How so? Can you test there that is a single expected allocation? I didn't see that yet.

I did try this but it didn't work:
```julia
julia> using FixedPointDecimals

julia> const WFD2 = FixedDecimal{Int128, 2}
FixedDecimal{Int128, 2}

julia> @check_allocs f(x) = convert(WFD2, x)
f (generic function with 2 methods)

julia> f(-1)
ERROR: ReadOnlyMemoryError()
```

nickrobinson251: > Can you test there that is a single expected allocation?


```suggestion
    @test @allocations(convert(UWFD2, -1)) == 1
```

Drvi: It seems to work on 1.9:
```julia
julia> using AllocCheck

julia> @check_allocs myconv(::Type{T}, x) where T = convert(T, x)
myconv (generic function with 1 method)

julia> using FixedPointDecimals

julia> const UWFD2 = FixedDecimal{UInt128, 2}
FixedDecimal{UInt128, 2}

julia> myconv(UWFD2, -1) # no error
FixedDecimal{UInt128,2}(1125651476239148520299089100.85)

julia> b = big(1)
1

julia> myconv(UWFD2, b)
ERROR: @check_alloc function encountered 6 errors (6 allocations / 0 dynamic dispatches).
```
Not sure about the `ReadOnlyMemoryError` I get on 1.10

NHDaly: aha that's right, i forgot this one throws; i wasn't looking very closely this morning. Thanks. I'm just not going to test this since it's the exception case. I'll just delete this line

NHDaly: But yeah you're right, if this was working, it would be a good way to test it!

Drvi: 👍

PR #88, opened 2023-12-19
Don't throw InexactError for comparisons between FDs of different types and/or Integers

### LINKED ISSUE #82: `==` should not leak promotion failure
```
julia> x = FixedDecimal{Int128, 33}(10000)
FixedDecimal{Int128,33}(10000.000000000000000000000000000000000)

julia> y = FixedDecimal{Int128, 38}(0)
FixedDecimal{Int128,38}(0.00000000000000000000000000000000000000)

julia> x == y
ERROR: InexactError: Int128(1000000000000000000000000000000000000000000)
Stacktrace:
 [1] Int128(x::BigInt)
   @ Base.GMP ./gmp.jl:373
 [2] convert
   @ ~/.julia/packages/FixedPointDecimals/Eseyj/src/FixedPointDecimals.jl:313 [inlined]
 [3] _promote
   @ ./promotion.jl:358 [inlined]
 [4] promote
   @ ./promotion.jl:381 [inlined]
 [5] ==(x::FixedDecimal{Int128, 33}, y::FixedDecimal{Int128, 38})
   @ Base ./promotion.jl:449
 [6] top-level scope
   @ REPL[5]:1
```

This should be not equal rather than `InexactError` (an implementation detail). Checking representability in `==` should solve this.

NHDaly:
Fixes #82.

Before:
```julia
FD{Int8, 0}(2) != FD{Int8, 2}(1)  # Throws because FD{Int8,2}(2) doesn't fit
2 != FD{Int8, 2}(1)  # Throws because FD{Int8,2}(2) doesn't fit
2 >= FD{Int8, 2}(1)  # Throws because FD{Int8,2}(2) doesn't fit
```
After, these all return true. :)

The idea in this PR is to check whether the values would fit in eachother's types, and if not, that's your answer, rather than throwing an InexactError if they fail to promote.


NHDaly:
Note that this PR _doesn't_ address InexactError caused by promotion during arithmetic, e.g.:
```julia
julia> FixedDecimal{Int8,2}(1) + 2  # Same here: 2 is promoted to FD{Int8,2}(2)
ERROR: InexactError: convert(FixedDecimal{Int8, 2}, 2)
```

I think that's okay for now. If we fixed it, this would either throw an OverflowError anyway if you used checked arithmetic, or it would wrap, using the new semantics. We should probably add support for this, but at least at RAI we're using checked arithmetic for everything so this is going to throw either way and so it seems less urgent to fix it to me.

src/FixedPointDecimals.jl
```diff
@@ -440,12 +440,12 @@ end
 
 function Base.checked_neg(x::T) where {T<:FD}
     r = -x
-    (x<0) & (r<0) && Base.Checked.throw_overflowerr_negation(x)
+    (x.i<0) & (r<0) && Base.Checked.throw_overflowerr_negation(x)
```
mcmcgrath13, line 443: what's happening here where sometimes we check the `.i` and sometimes we don't? (maybe add a comment?)

NHDaly: I think we should always look at th `.i`, thanks. I noticed it while fixing something else, and missed the other case. Lemme change that too

src/FixedPointDecimals.jl
```diff
@@ -537,6 +537,88 @@ Base.:(==)(x::T, y::T) where {T <: FD} = x.i == y.i
 Base.:(<)(x::T, y::T) where {T <: FD} = x.i < y.i
 Base.:(<=)(x::T, y::T) where {T <: FD} = x.i <= y.i
 
+# Avoid promotions when possible to avoid throwing InexactError
+for comp_op in (:(==), :(<), :(<=))
```
mcmcgrath13, line 541: does this overload the operators above? or do those still hold when it's the same type of fixed decimal?

src/FixedPointDecimals.jl
NHDaly: done, added comment

src/FixedPointDecimals.jl
```diff
@@ -537,6 +537,88 @@ Base.:(==)(x::T, y::T) where {T <: FD} = x.i == y.i
...
+        if f == 0
+            # If the precisions are the same, even if the types are different, we can
+            # compare the integer values directly. e.g. Int8(2) == Int16(2) is true.
+            return $comp_op(x, y.i)
+        else
+            # If x is too big to fit in T, then we know already that it's bigger than y,
+            # so not equal and not less than.
+            if !(x isa T) && x > typemax(T)
+                return false
+            end
```
mcmcgrath13, line 585: do we also need to check it's lower than the type min?

src/FixedPointDecimals.jl
NHDaly: done

src/FixedPointDecimals.jl
NHDaly: Yes, we do! 💡 great catch. I wonder why I didn't catch that in the last set of tests I added.. 🤔

NHDaly: Aha, i've got tests that capture this now. Thanks Mary, good catch!
```julia
    @test typemax(Int16) > FD{Int8, 0}(-1) > typemin(Int16)
    @test typemax(Int16) > FD{Int8, 2}(-1) > typemin(Int16)
    @test typemin(Int16) < FD{Int8, 0}(-1) < typemax(Int16)
    @test typemin(Int16) < FD{Int8, 2}(-1) < typemax(Int16)
```

NHDaly: Thanks, fixed!

NHDaly:
Thanks for the quick review! Ready for another round 😊

mcmcgrath13 (approved):
❤️ thank you!! Can you just make sure we have full test coverage before merging?

NHDaly:
Thanks Mary. 100% diff coverage now. (That's a really helpful tool.)

Here I go! :) I'm gonna also mark this as a patch version bump, since it's a strict improvement.

PR #89, opened 2023-12-20
Fix bug in previous PR for comparisons that underflow (large negatives)

NHDaly:
In these cases, we were previously returning the wrong result. 😓 
```julia
julia> @test typemin(Int8) < FD{Int8, 2}(-1) < typemax(Int8)
Test Failed at REPL[27]:1
  Expression: typemin(Int8) < FD{Int8, 2}(-1) < typemax(Int8)
   Evaluated: -128 < FixedDecimal{Int8,2}(-1.00) < 127

julia> @test typemin(FD{Int8, 1}) < typemin(FD{Int8,2})
Test Failed at REPL[34]:1
  Expression: typemin(FD{Int8, 1}) < typemin(FD{Int8, 2})
   Evaluated: FixedDecimal{Int8,1}(-12.8) < FixedDecimal{Int8,2}(-1.28)
```
Now those tests pass.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "FixedPointDecimals"
 uuid = "fb4d412d-6eee-574d-9565-ede6634db7b0"
 authors = ["Fengyang Wang <fengyang.wang.0@gmail.com>", "Curtis Vogt <curtis.vogt@gmail.com>"]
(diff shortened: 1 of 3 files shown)


mcmcgrath13 (approved):
thank you!!

PR #93, opened 2024-06-12
Use Int256 to reduce BigInts in FD operations.

NHDaly:
We do not here explicitly introduce support for FD{BitIntegers.Int256}, though that should work out of the box both before and after this PR.

Rather, this PR _uses_ a (U)Int256 under the hood to prevent allocations from Int128 widening to BigInt in FD operations.
Unfortunately, `rem` and `mod` on BitIntegers.Int256 still fall-back to a BigInt ([see the note here](https://github.com/rfourquet/BitIntegers.jl/blob/deb17db47649cc59b49a3dfd29d74efa311590d7/src/BitIntegers.jl#L410-L414)), so this doesn't completely eliminate the BigInt allocs. But it does reduce them.

------

This is a pretty small PR, but it should have a big impact on users of `FD{Int128}`.

Before:
```julia
julia> @btime fd * fd setup = (fd = FixedDecimal{Int128,3}(1.234))
  392.413 ns (24 allocations: 464 bytes)
FixedDecimal{Int128,3}(1.523)
```
After:
```julia
julia> @btime fd * fd setup = (fd = FixedDecimal{Int128,3}(1.234))
  213.039 ns (12 allocations: 240 bytes)
FixedDecimal{Int128,3}(1.523)
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "FixedPointDecimals"
 uuid = "fb4d412d-6eee-574d-9565-ede6634db7b0"
 authors = ["Fengyang Wang <fengyang.wang.0@gmail.com>", "Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "0.5.2"
+version = "0.5.3"
 
 [deps]
 BitIntegers = "c3b6d118-76ef-56ca-8cc7-ebb389d030a1"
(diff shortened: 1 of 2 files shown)


NHDaly:
I just realized this reimplements https://github.com/RelationalAI-oss/FixedPointDecimals.jl/pull/7, from 6 years ago (😳) which @TotalVerb had already reviewed. @TotalVerb you may want to do one more pass over this.

Drvi (approved):
LGTM! Thanks

PR #94, opened 2024-06-13
Finish removing the BigInts from * for FD{Int128}!

NHDaly:
Finally implements the fast-multiplication optimization from https://github.com/JuliaMath/FixedPointDecimals.jl/pull/45! :)

This PR optimizes multiplication for `FixedDecimal{Int64}` and `FixedDecimal{Int128}`. In the process, we also undid an earlier optimization which is no longer needed after julia 1.8, and that makes multiplication about 2x fast for the smaller int types as well! 🎉 

This is a follow-up to
https://github.com/JuliaMath/FixedPointDecimals.jl/pull/93, which introduces an Int256 type for widemul. However after that PR, the fldmod still required 2 BigInt allocations.

Now, this PR uses a custom implementation of the LLVM div-by-const optimization for (U)Int128 and for (U)Int256, which briefly widens to Int512 (😅), to perform the fldmod by the constant 10^f coefficient.

After this PR, FD multiply performance scales linear with the number of bits. FD{Int128} has no allocations, and is only 2x slower than 64-bit. :) And it makes all other multiplications ~2x faster.

master:
```julia
julia> using FixedPointDecimals, BenchmarkTools

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{Int16,3}(1.234))
  83.750 μs (0 allocations: 0 bytes)
FixedDecimal{Int16,3}(0.000)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{Int32,3}(1.234))
  84.916 μs (0 allocations: 0 bytes)
FixedDecimal{Int32,3}(1700943.280)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{Int64,3}(1.234))
  249.083 μs (0 allocations: 0 bytes)
FixedDecimal{Int64,3}(4230510070790917.029)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{Int128,3}(1.234))
  4.660 ms (248829 allocations: 4.70 MiB)
FixedDecimal{Int128,3}(-66726338547984585007169386718143307.324)

# unsigned

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{UInt16,3}(1.234))
  56.791 μs (0 allocations: 0 bytes)
FixedDecimal{UInt16,3}(0.000)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{UInt32,3}(1.234))
  60.000 μs (0 allocations: 0 bytes)
FixedDecimal{UInt32,3}(4191932.283)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{UInt64,3}(1.234))
  172.958 μs (0 allocations: 0 bytes)
FixedDecimal{UInt64,3}(16576189118051436.703)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{UInt128,3}(1.234))
  5.408 ms (308621 allocations: 6.14 MiB)
FixedDecimal{UInt128,3}(303384805088638153637410092093845905.434)
```
After PR #93:
```julia
julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{Int16,3}(1.234))
  83.750 μs (0 allocations: 0 bytes)
FixedDecimal{Int16,3}(0.000)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{Int32,3}(1.234))
  85.000 μs (0 allocations: 0 bytes)
FixedDecimal{Int32,3}(1700943.280)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{Int64,3}(1.234))
  248.958 μs (0 allocations: 0 bytes)
FixedDecimal{Int64,3}(4230510070790917.029)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{Int128,3}(1.234))
  4.673 ms (160798 allocations: 3.22 MiB)
FixedDecimal{Int128,3}(-66726338547984585007169386718143307.324)

# unsigned

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{UInt16,3}(1.234))
  56.791 μs (0 allocations: 0 bytes)
FixedDecimal{UInt16,3}(0.000)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{UInt32,3}(1.234))
  60.041 μs (0 allocations: 0 bytes)
FixedDecimal{UInt32,3}(4191932.283)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{UInt64,3}(1.234))
  173.000 μs (0 allocations: 0 bytes)
FixedDecimal{UInt64,3}(16576189118051436.703)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{UInt128,3}(1.234))
  4.750 ms (190708 allocations: 4.82 MiB)
FixedDecimal{UInt128,3}(303384805088638153637410092093845905.434)

```
After this PR:
```julia
julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{Int16,3}(1.234))
  48.458 μs (0 allocations: 0 bytes)
FixedDecimal{Int16,3}(0.000)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{Int32,3}(1.234))
  57.000 μs (0 allocations: 0 bytes)
FixedDecimal{Int32,3}(1700943.280)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{Int64,3}(1.234))
  90.708 μs (0 allocations: 0 bytes)
FixedDecimal{Int64,3}(4230510070790917.029)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{Int128,3}(1.234))
  180.125 μs (0 allocations: 0 bytes)
FixedDecimal{Int128,3}(-66726338547984585007169386718143307.324)

# unsigned

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{UInt16,3}(1.234))
  42.708 μs (0 allocations: 0 bytes)
FixedDecimal{UInt16,3}(0.000)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{UInt32,3}(1.234))
  51.250 μs (0 allocations: 0 bytes)
FixedDecimal{UInt32,3}(4191932.283)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{UInt64,3}(1.234))
  80.042 μs (0 allocations: 0 bytes)
FixedDecimal{UInt64,3}(16576189118051436.703)

julia> @btime for _ in 1:10000 fd = fd * fd end setup = (fd = FixedDecimal{UInt128,3}(1.234))
  162.417 μs (0 allocations: 0 bytes)
FixedDecimal{UInt128,3}(303384805088638153637410092093845905.434)
```


NHDaly:
Huzzah! I finally feel like we can bring the ideas from https://github.com/JuliaMath/FixedPointDecimals.jl/pull/45 to this package, and it's still valuable. It turns out that LLVM has already applied this optimization automatically for Int128 div by const (thus * for `FixedDecimal{Int64}` is already fast). But this PR adds support for `FixedDecimal{Int128}`, which needed a custom function for Int256 div by const. :)

See here, for my note on FixedDecimal{Int64} support:
https://github.com/JuliaMath/FixedPointDecimals.jl/pull/45#issuecomment-2163836966

NHDaly:
Now that there's julia's effects system, this doesn't need `@pure`, so it's actually possible to do this! 
And now that we've introduced BitIntegers.jl, the code is way simpler. 😊 woohoo!

test/runtests.jl
```diff
@@ -10,4 +10,12 @@ include(joinpath(pkg_path, "test", "utils.jl"))
...
     include("FixedDecimal.jl")
-end  # global testset
+end
+
+
+if isdefined(Base, Symbol("@assume_effects"))
+@testset "FixedPointDecimals" begin
+    include("fldmod-by-const_tests.jl")
+end
+end
```
omus, line 20: Why not include this in the existing FixedPointDecimals testset?
```suggestion
    isdefined(Base, Symbol("@assume_effects")) && include("fldmod-by-const_tests.jl")
end
```

src/FixedPointDecimals.jl
```diff
@@ -36,9 +36,22 @@ export checked_abs, checked_add, checked_cld, checked_div, checked_fld,
 
 using Base: decompose, BitInteger
 
-import BitIntegers  # For 128-bit _widemul / _widen
+using BitIntegers: BitIntegers, UInt256, Int256
```
omus, line 39: Why not import the other used types?
```suggestion
using BitIntegers: BitIntegers, Int256, Int512, UInt256, UInt512
```

src/fldmod-by-const.jl
```diff
@@ -0,0 +1,98 @@
...
+
+function mul_hi(x::T, y::T) where T
+    xy = _widemul(x, y)  # support Int256 -> Int512 (!!)
+    (xy >> nbits(T)) % T
+end
+
+# Annoyingly, Unsigned(T) isn't defined for BitIntegers types:
+# https://github.com/rfourquet/BitIntegers.jl/pull/2
+_unsigned(x) = unsigned(x)
+_unsigned(::Type{Int256}) = UInt256
```
omus, line 96: Is there a missing test?
```suggestion
_unsigned(::Type{Int256}) = UInt256
_unsigned(::Type{Int512}) = UInt512
```

test/runtests.jl
NHDaly: Oh, oops, I meant to rename the test-set to `fldmod-by-const_tests`.

🤷 i dunno, i guess it seems nice to have a separate test-set per file, so that you can group the failures? but you're right that we probably _also_ want an outer testset so that a failure in one doesn't abort the rest of the tests.. annoying.

I'll take your change, i think that's preferable.

src/fldmod-by-const.jl
NHDaly: No, the built-in `unsigned(T)` already worked for the UInt256, and I think this isn't needed for UInt512, since we are only widening to 512 _after_ we call unsigned.

```julia
julia> unsigned(FixedPointDecimals.UInt256)
BitIntegers.UInt256

julia> unsigned(FixedPointDecimals.Int256)
ERROR: MethodError: no method matching rem(::Type{BitIntegers.Int256}, ::Type{BitIntegers.UInt256})
```

Lemme add a comment to explain that.

NHDaly: Done. Have another look there?

src/FixedPointDecimals.jl
NHDaly: Seems reasonable. The (U)Int256 were used much more heavily than (U)Int512. So I was inclined to refer to them directly through the qualified name. But I think your suggestion is better.

NHDaly:
Update 1: It turned out that (of course) `*` is too fast for `@btime` to get a good measurement of on its own. I've adjusted the benchmark to perform repeated multiplies, and the true performance is now included in the original comment.

Update 2: It turns out that this is actually faster for _all_ int types (>2x faster for FD{Int64}), even though LLVM was doing a version of this optimization on its own. I'm honestly pretty surprised by that... From what I can tell, for FD{Int32}, the code is _identical_ except for a single instruction changes from comparing `hi` to comparing `gt`:
Screenshot 2024-06-13 at 2 10 20 PM
And yet somehow it's 25% faster (80µs -> 60µs).

NHDaly:
Aha, the native code difference comes from the whole benchmark function. Before, the multiply was outlined, and called through a function call.
Now, it's inlined, and i think the whole loop is constant-folded away, so it's actually sort of cheating. 😅 
But still, allowing that optimization seems like a positive thing!

src/FixedPointDecimals.jl
```diff
@@ -129,8 +142,10 @@ _widemul(x::Unsigned,y::Signed) = signed(_widen(x)) * _widen(y)
 
 # Custom widen implementation to avoid the cost of widening to BigInt.
 # FD{Int128} operations should widen to 256 bits internally, rather than to a BigInt.
-_widen(::Type{Int128}) = BitIntegers.Int256
-_widen(::Type{UInt128}) = BitIntegers.UInt256
+_widen(::Type{Int128}) = Int256
+_widen(::Type{UInt128}) = UInt256
+_widen(::Type{Int256}) = BitIntegers.Int512
+_widen(::Type{UInt256}) = BitIntegers.UInt512
```
omus, line 148:
```suggestion
_widen(::Type{Int256}) = Int512
_widen(::Type{UInt256}) = UInt512
```

omus (commented):
Did another surface level review. I can do a full review if no one else will. I will need to block off some time though as there's a bit going on in this PR.

NHDaly:
I _think_ Tomas should be able to review when he's back from vacation. He's out til thursday. :)

NHDaly:
Okay, after reviewing and integrating all of Tomas' changes, and cleaning things up, i think this is a great improvement and is ready to go! :) Tomas and I are going to look through it one more time, but otherwise I think this is fully ready for review. Thanks!

NHDaly:
I also updated the PR comment with the final perf numbers, which look 👌

EdsterG:
By the way, `div`/`mod` allocations are fixed in the most recent version of BitIntegers when running on 1.11+, see [#48](https://github.com/rfourquet/BitIntegers.jl/pull/48).

NHDaly:
Oh wow, very nice! Thanks @EdsterG. 💪 

This should still help quite a bit, but the baseline for comparison would be better in 1.11, then! 💪

src/fldmod-by-const.jl
```diff
@@ -0,0 +1,131 @@
...
+    elseif C <= 0
+        throw(DomainError("C must be > 0"))
+    end
+    # Calculate the magic number and shift amount, based on Hacker's Delight, Chapter 10.
+    magic_number, shift = magicg(typemax(T), C)
+
+    out = _widemul(promote(x, magic_number)...)
+    out >>= shift
+    # Adding one as implied by formula (1b) in Hacker's delight, Chapter 10-4.
+    return (out % T) + (x < zero(T))
```
NHDaly:
```suggestion
    # Add one if x was negative as implied by formula (1b) in Hacker's delight, 
    # Chapter 10-4. (Note that of course if x is Unsigned, this compiles away.)
    return (out % T) + (x < zero(T))
```

Drvi:
Ok, so I've reviewed the code and I concluded that it is correct. 

I was mainly concerned about whether should or should not add 1 for negative `x` in the last line of `div_by_const`. I later realized that, as Nathan pointed out, the +1 is very much intended and explicitly relied on in the associated proofs in Hacker's Delight. The high level reason for it is that shifts implement flooring division while we were interested in division towards zero, so for negative values, if the division is not already exact, we need to add one to compensate for the flooring. I've spent some time proving that we should always add 1 in this code path (i.e. that we don't divide evenly and that we need to compensate). I've added the proof to the body of the function as a comment, but now I think that it is not useful, since it proves a subset of what the relevant section in HD proves already (i.e. HD proves the multiplication and shift corresponds to `ceil` for negative numbers if we add 1, and `floor` for positive numbers, while I prove that given the multitplication and shift, we need to add the 1 for negative numbers).

I've also added JET tests and manually examined generated typed IR for all possible multiplication methods to verify that none of those mentioned any errors or unreachable branches. 

I've also run the exhaustive tests for `Int16`, all of them passed.

NHDaly (commented):
Alright, finally ready to go! I've moved @Drvi's last comment out to below the function, to make it easier to read. But this is finally ready to merge. I've also confirmed the performance impacts again locally.

Project.toml
NHDaly:
```suggestion
version = "0.6.5"
```

NHDaly: Bump to 0.6.5

PR #102, opened 2025-01-16
Add with_overflow for add, sub, and mul

mbarbar:
Mostly pulled from `checked_x` which are then made to use these. The tests are copies of the existing tests for `checked_x` with the addition of a few tests exercising non-overflowing cases.


NHDaly (approved):
Lovely! :) Thanks @mbarbar! LGTM

PR #104, opened 2025-01-23
Bck aqua clean

kuszmaul:
Make `Aqua.test_all(FixedPointDecimal)` pass.

Increment the patch level.

Add compat values for the extras in Project.toml

Import Base symbols that are reexported.

Define extra methods to disambiguate.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -8,8 +8,10 @@ BitIntegers = "c3b6d118-76ef-56ca-8cc7-ebb389d030a1"
 Parsers = "69de0a69-1ddd-5017-9359-2bf0b02dc9f0"
 
 [compat]
-Parsers = "2.7"
 BitIntegers = "0.3.1"
+Parsers = "2.7"
+Printf = ">=1.6"
+Test = ">=1.6"
 julia = "1.6"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "FixedPointDecimals"
 uuid = "fb4d412d-6eee-574d-9565-ede6634db7b0"
 authors = ["Fengyang Wang <fengyang.wang.0@gmail.com>", "Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "0.5.4"
+version = "0.5.5"
 
 [deps]
 BitIntegers = "c3b6d118-76ef-56ca-8cc7-ebb389d030a1"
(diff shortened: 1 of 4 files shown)


Project.toml
```diff
@@ -1,15 +1,17 @@
...
 [deps]
 BitIntegers = "c3b6d118-76ef-56ca-8cc7-ebb389d030a1"
 Parsers = "69de0a69-1ddd-5017-9359-2bf0b02dc9f0"
 
 [compat]
-Parsers = "2.7"
 BitIntegers = "0.3.1"
+Parsers = "2.7"
+Printf = ">=1.6"
+Test = ">=1.6"
```
nickrobinson251, line 14: As written these compat entries promise to be compatible with all future versions of Printf and Test, including e.g. a v2 with many breaking changes

I think we probably only want to declare compatibility with those that are SemVer compatible with 1.6, which we can do via:

```suggestion
Printf = "1.6"
Test = "1.6"
```

https://pkgdocs.julialang.org/v1/compatibility/#Version-specifier-format

nickrobinson251 (approved):
Shall we add `Aqua.test_all(FixedPointDecimal)` as a test in this repo?

kuszmaul:
> Shall we add `Aqua.test_all(FixedPointDecimal)` as a test in this repo?

I'll do that

kuszmaul:
I added Aqua tests and also unit tests for the newly-defined methods.  I think this is ready to merge.  I don't have write access to this repo.

test/aqua_test.jl
```diff
@@ -0,0 +1,4 @@
+@testitem "Aqua" begin
```
nickrobinson251, line 1: This package doesn't use ReTestItems.jl, just Test.jl
```suggestion
@testset "Aqua" begin
```

test/aqua_test.jl
```diff
@@ -0,0 +1,4 @@
...
+    using Aqua
+    Aqua.test_all(FixedPointDecimal)
+end
```
nickrobinson251, line 4: This new file will need adding as an `include` to `test/runtest.jl`
https://github.com/JuliaMath/FixedPointDecimals.jl/blob/e36a42f04a1105a15e582a1e0adee29a159d7293/test/runtests.jl#L11-L13

nickrobinson251:
test failures of `nightly` julia look related to https://github.com/JuliaLang/julia/issues/57064

PR #105, opened 2025-01-25
Remove promoting with_overflow overloads

mbarbar:
This is a bit odd because it can cause an overflow and not be handled. It is also not the interface offered by the standard library.
```
julia> methods(Base.Checked.add_with_overflow)
# 4 methods for generic function "add_with_overflow" from Base.Checked:
 [1] add_with_overflow(a::BigInt, b::BigInt)
     @ Base.GMP gmp.jl:838
 [2] add_with_overflow(x::Bool, y::Bool)
     @ checked.jl:143
 [3] add_with_overflow(x::T, y::T) where T<:Union{UInt128, UInt16, UInt32, UInt64, UInt8}
     @ checked.jl:142
 [4] add_with_overflow(x::T, y::T) where T<:Union{Int128, Int16, Int32, Int64, Int8}
     @ checked.jl:141

```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "FixedPointDecimals"
 uuid = "fb4d412d-6eee-574d-9565-ede6634db7b0"
 authors = ["Fengyang Wang <fengyang.wang.0@gmail.com>", "Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "0.6"
+version = "0.6.0"
 
 [deps]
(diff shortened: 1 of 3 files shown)


NHDaly (commented):
Ah, interesting. Good call!

Can you please also bump the version number in the Project.toml to `0.6`, since this is a backwards-incompatible change?

NHDaly (approved):

NHDaly (approved):

PR #106, opened 2025-01-28
Add div_with_overflow

mbarbar:


src/FixedPointDecimals.jl
```diff
@@ -441,6 +441,29 @@ Base.Checked.mul_with_overflow(x::FD, y::FD) = Base.Checked.mul_with_overflow(pr
...
+Return the result of div (wrapping on overflow/underflow) and a boolean indicating whether
+overflow/underflow did in fact happen. Throws on divide-by-zero.
+"""
+function div_with_overflow(x::FD{T,f}, y::FD{T,f}) where {T<:Integer,f}
+    C = coefficient(FD{T, f})
+    # This case will break the div call below.
+    if x.i == typemin(T) && y.i == -1
+        # To perform the div and overflow means reaching the max and adding 1, so typemin.
+        return (typemin(FD{T,f}), true)
+    end
```
NHDaly, line 459: yeah, interesting! is typemin/-1 overflow or is it the same kind of "undefined" error as divide-by-0?

🤔 i think your choice is right. 👍 typemin/-1 is just overflow -- with a bigger integer type it would've succeeded. Whereas divide by 0 is a truly undefined point in the function. 👍 I like this choice

src/FixedPointDecimals.jl
```diff
@@ -441,6 +441,29 @@ Base.Checked.mul_with_overflow(x::FD, y::FD) = Base.Checked.mul_with_overflow(pr
...
+    # Note: The div() will throw for divide-by-zero, that's not an overflow.
+    v, b = Base.Checked.mul_with_overflow(C, div(x.i, y.i))
+    # div the result by 1 to drop the decimal part which could come about because
+    # of the wrapping in mul_with_overflow.
+    return (div(reinterpret(FD{T,f}, v), 1), b)
```
NHDaly, line 464: I don't think it's worth adding this extra `div` - division is very expensive.

I think if your function returns `true`, the result no longer needs to hold to the constraints of the function. As a simple example:
```julia
julia> Base.Checked.add_with_overflow(Int8(127), Int8(1))
(-128, true)
```
obviously 127 + 1 is not -128. So I _think_ you should just let this wrap and return the true, wrapped value?

src/FixedPointDecimals.jl
```diff
@@ -441,6 +441,29 @@ Base.Checked.mul_with_overflow(x::FD, y::FD) = Base.Checked.mul_with_overflow(pr
...
+    C = coefficient(FD{T, f})
+    # This case will break the div call below.
+    if x.i == typemin(T) && y.i == -1
+        # To perform the div and overflow means reaching the max and adding 1, so typemin.
+        return (typemin(FD{T,f}), true)
+    end
+    # Note: The div() will throw for divide-by-zero, that's not an overflow.
+    v, b = Base.Checked.mul_with_overflow(C, div(x.i, y.i))
+    # div the result by 1 to drop the decimal part which could come about because
+    # of the wrapping in mul_with_overflow.
```
NHDaly, line 463:
```suggestion
```

src/FixedPointDecimals.jl
mbarbar: Yeah, in fact the error thrown by the standard library is an OverflowError.

src/FixedPointDecimals.jl
```diff
@@ -441,6 +441,29 @@ Base.Checked.mul_with_overflow(x::FD, y::FD) = Base.Checked.mul_with_overflow(pr
...
+
+Return the result of div (wrapping on overflow/underflow) and a boolean indicating whether
+overflow/underflow did in fact happen. Throws a DivideError on divide-by-zero.
+"""
+function div_with_overflow(x::FD{T,f}, y::FD{T,f}) where {T<:Integer,f}
+    C = coefficient(FD{T, f})
+    # This case will break the div call below.
+    if x.i == typemin(T) && y.i == -1
+        # To perform the div and overflow means reaching the max and adding 1, so typemin.
+        return (typemin(FD{T,f}), true)
```
NHDaly, line 458:
```suggestion
        return (x, true)
```
i think this would be the same code anyway, but it makes it a bit simpler to read

NHDaly (approved):
LGTM!

PR #108, opened 2025-01-31
Fix world age warnings from accessing binding before definiton 

### LINKED ISSUE #107: Warnings in Julia v1.12.0-DEV `Detected access to binding in a world prior to its definition world`
```julia
(FixedPointDecimals) pkg> instantiate
    Updating registry at `~/.julia/registries/General.toml`
    Updating `~/repos/FixedPointDecimals.jl/Project.toml`
  [c3b6d118] + BitIntegers v0.3.2
  [69de0a69] + Parsers v2.8.1
    Updating `~/repos/FixedPointDecimals.jl/Manifest.toml`
  [c3b6d118] + BitIntegers v0.3.2
  [69de0a69] + Parsers v2.8.1
  [aea7be01] + PrecompileTools v1.2.1
  [21216c6a] + Preferences v1.4.3
  [ade2ca70] + Dates v1.11.0
  [de0858da] + Printf v1.11.0
  [9a3f8284] + Random v1.11.0
  [ea8e919c] + SHA v0.7.0
  [fa267f1f] + TOML v1.0.3
  [cf7118a7] + UUIDs v1.11.0
  [4ec0a83e] + Unicode v1.11.0
Precompiling packages finished.
  1 dependency successfully precompiled in 2 seconds. 9 already precompiled.
  1 dependency had output during precompilation:
┌ FixedPointDecimals
│  WARNING: Detected access to binding `FixedPointDecimals.truncmul` in a world prior to its definition world.
│    Julia 1.12 has introduced more strict world age semantics for global bindings.
│    !!! This code may malfunction under Revise.
│    !!! This code will error in future versions of Julia.
│  Hint: Add an appropriate `invokelatest` around the access to this binding.
│  WARNING: Detected access to binding `FixedPointDecimals.floormul` in a world prior to its definition world.
│    Julia 1.12 has introduced more strict world age semantics for global bindings.
│    !!! This code may malfunction under Revise.
│    !!! This code will error in future versions of Julia.
│  Hint: Add an appropriate `invokelatest` around the access to this binding.
│  WARNING: Detected access to binding `FixedPointDecimals.ceilmul` in a world prior to its definition world.
│    Julia 1.12 has introduced more strict world age semantics for global bindings.
│    !!! This code may malfunction under Revise.
│    !!! This code will error in future versions of Julia.
│  Hint: Add an appropriate `invokelatest` around the access to this binding.
└

julia> VERSION
v"1.12.0-DEV.1954"

```

nickrobinson251:
Close #107

This warning is only present starting in Julia 1.12

Presumably the issue is with the docstring resolving the binding `fnname` before it's actually been defined. I don't fully understand the problem but this fixes it 🤷‍♂️

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -47,11 +47,12 @@ const FMAFloat = Union{Float16, Float32, Float64, BigFloat}
 
 for fn in [:trunc, :floor, :ceil]
     fnname = Symbol(fn, "mul")
+    fnname_str = String(fnname)
     opp_fn = fn == :floor ? :ceil : :floor
 
     @eval begin
         @doc """
-            $($fnname)(I, x, y) :: I
+            $($fnname_str)(I, x, y) :: I
 
         Compute `$($fn)(I, x * y)`, returning the result as type `I`. For
         floating point values, this function can be more accurate than
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "FixedPointDecimals"
 uuid = "fb4d412d-6eee-574d-9565-ede6634db7b0"
 authors = ["Fengyang Wang <fengyang.wang.0@gmail.com>", "Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "0.6.0"
+version = "0.6.1"
 
 [deps]
 BitIntegers = "c3b6d118-76ef-56ca-8cc7-ebb389d030a1"


src/FixedPointDecimals.jl
```diff
@@ -47,11 +47,12 @@ const FMAFloat = Union{Float16, Float32, Float64, BigFloat}
 
 for fn in [:trunc, :floor, :ceil]
     fnname = Symbol(fn, "mul")
+    fnname_str = String(fnname)
     opp_fn = fn == :floor ? :ceil : :floor
 
     @eval begin
         @doc """
-            $($fnname)(I, x, y) :: I
+            $($fnname_str)(I, x, y) :: I
```
NHDaly, line 55: Ah, nice, thanks! Good find! It took me a minute of staring, but i think i've got an explanation:
It's easy to forget, but interpolating a symbol into an expression creates an _identifier_, not a symbol.
So this was generating the following string code:
```
"  $(trunc_mul)(I, x, y) :: I "
```
and as you point out, the `trunc_mul` function doesn't exist yet, so this should fail.

What we _meant_ to be putting in there was the name of the function, as you've done, so your change looks great. 👍 
The other thing we could have done is to generate what this code **meant** to be generating, which is this:
```julia
julia> "$(:foo)(I, x, y) :: I"
"foo(I, x, y) :: I"
```
That is, putting the symbol _itself_ into the code, rather than using the symbol to put an identifier in the code.

The syntax to do that is to wrap the value in a QuoteNode, whose sole purpose it to allow putting symbols or expressions _as values_ into generated code:
```julia
            $($(Meta.quot(fnname)))(I, x, y) :: I
            $($(QuoteNode(fnname)))(I, x, y) :: I
```
(either of those options are identical)

NHDaly: but anyway, i like what you've done - it's very straightforward.

NHDaly (approved):

src/FixedPointDecimals.jl
NHDaly: in particular, the quote suggestion raises the question "how does a symbol stringify in interpolation? Does it print as `foo` or `:foo`?" so converting to a string is more clear 👍

PR #109, opened 2025-02-06
Add rdiv_with_overflow and fld_with_overflow

mbarbar:


src/FixedPointDecimals.jl
```diff
@@ -454,6 +454,78 @@ function div_with_overflow(x::FD{T,f}, y::FD{T,f}) where {T<:Integer,f}
...
+The overflow protection may impose a perceptible performance penalty.
+
+See also:
+- `Base.checked_fld`.
+"""
+function fld_with_overflow(x::FD{T,f}, y::FD{T,f}) where {T<:Integer,f}
+    C = coefficient(FD{T, f})
+    # This case will break the fld call below. This can only happen when f is 0 as y.i
+    # cannot be -1 otherwise.
+    if x.i == typemin(T) && y.i == -1
```
mbarbar, line 474: This is an introduced `typemin`, completely new.

src/FixedPointDecimals.jl
```diff
@@ -454,6 +454,78 @@ function div_with_overflow(x::FD{T,f}, y::FD{T,f}) where {T<:Integer,f}
...
+        # typemin / -1 for signed integers wraps, giving typemin (x) again.
+        return (x, true)
+    end
+
+    quotient, remainder = fldmod(x.i, y)
+    # It is impossible for both the quotient to be typemax/typemin AND remainder to be
+    # non-zero because y is an integer. Thus the following call cannot overflow.
+    v = _round_to_nearest(quotient, remainder, y)
+    return (reinterpret(FD{T, f}, v), false)
+end
```
NHDaly, line 527: As you did with our other `_with_overflow` operations, should we simply remove these promoting variants, and say that if you want to do rdiv_with_overflow, you must first promote the arguments, and check for overflow during promotion?

NHDaly: I suppose that FD / FD is less efficient than FD / Int though..... 🤔 interesting. What do you think?

mbarbar: With these variants specifically, the issue becomes like the comment says (this was on the original `checked_div`) that this expands the possible results. I now realise multiplication has the same issue, though the `checked_mul(::Integer, ::FD)` is a naive promoting overload.
```
Base.checked_mul(x, y::FD) = Base.checked_mul(promote(x, y)...)
```

For example, with these you can do `rdiv_with_overflow(FD{Int8,2}(1), 2)`. If you try to promote, you have overflow.

I'm not certain what we should do here. Is there precedent for these kinds of things in Julia for us to follow?

In the context of Rel I don't know what is desired and I'm not certain anyone knows. Numbers and operators have yet to actually be defined.

mbarbar: Confirmed we need this in light of https://relationalai.slack.com/archives/C03G30AQX3K/p1739176044123089?thread_ts=1738990402.856679&cid=C03G30AQX3K

src/FixedPointDecimals.jl
```diff
@@ -454,6 +454,78 @@ function div_with_overflow(x::FD{T,f}, y::FD{T,f}) where {T<:Integer,f}
...
+The overflow protection may impose a perceptible performance penalty.
+
+See also:
+- `Base.checked_fld`.
+"""
+function fld_with_overflow(x::FD{T,f}, y::FD{T,f}) where {T<:Integer,f}
+    C = coefficient(FD{T, f})
+    # This case will break the fld call below. This can only happen when f is 0 as y.i
+    # cannot be -1 otherwise.
+    if T <: Signed && x.i == typemin(T) && y.i == -1
```
mbarbar, line 474: This is an introduced typemin, completely new, FYI.

NHDaly, line 474: 👀  I was looking at the `T<: Signed` here, and I wondered: Is there a bug in the `div_with_overflow` above, since we don't check for `T <: Signed` there?

But now that I think about it, if `T <: Unsigned`, then `y.i` will _never_ `== -1` :)
```julia
julia> typemax(UInt) == -1
false

julia> typemin(UInt) == -1
false

julia> UInt(-1)
ERROR: InexactError: check_top_bit(UInt64, -1)
```

I guess it's still nice to have the check there since it's semantically clearer. (And it might save julia some compile time to figure it out.)

It does seem like julia can compile this comparison away though even without the check 😎 Pretty rad.
```julia
julia> foo(x) = x == -1
foo (generic function with 1 method)

julia> @code_llvm debuginfo=:none foo(UInt(0))
define i8 @julia_foo_392(i64 zeroext %0) #0 {
top:
  ret i8 0
}
```

NHDaly: what do you mean by that?

NHDaly, line 474: Can you add the `f == 0` check here too? I don't know if julia would be able to infer that away?
```suggestion
    if T <: Signed && f == 0 && x.i == typemin(T) && y.i == -1
```

src/FixedPointDecimals.jl
NHDaly: Yeah, interesting. I think it makes sense to leave this, since it covers an otherwise unavailable computation, as you note. Very interesting, though!

test/FixedDecimal.jl
```diff
@@ -848,6 +874,16 @@ end
...
+            @test rdiv_with_overflow(FD{Int64,8}(5), FD{Int64,8}(2)) == (FD{Int64,8}(2.5), false)
+            @test rdiv_with_overflow(FD{Int64,8}(5), FD{Int64,8}(0.5)) == (FD{Int64,8}(10), false)
+            @test rdiv_with_overflow(FD{Int128,0}(20000), Int32(5000)) == (FD{Int128,0}(4), false)
+
+            @test fld_with_overflow(typemax(FD{Int128,38}), FD{Int128,38}(1)) == (FD{Int128,38}(1), false)
+            @test fld_with_overflow(FD{Int64,8}(20.5), FD{Int64,8}(2.1)) == (FD{Int64,8}(9), false)
+            @test fld_with_overflow(FD{Int8,0}(-5), FD{Int8,0}(-1)) == (FD{Int8,0}(5), false)
+            @test fld_with_overflow(FD{Int8,2}(0.99), FD{Int8,2}(0.5)) == (FD{Int8,2}(1), false)
         end
     end
```
NHDaly, line 888: thanks - all these tests are really nice. :)

NHDaly (approved):
LGTM thanks! Can you speak to some of my last comments then we can merge? :)

src/FixedPointDecimals.jl
mbarbar: Just wrt our discussion on typemin and BigInt.

mbarbar: Oof yeah my bad. Good point on the Unsigned disctinction. I've add the <: Signed check, do you have a preference of with/without?

mbarbar: I don't know what I was thinking but that comment was wrong. I guess I was looking at the integral part of `y`. `y.i` can be -1 when you have e.g. `FixedDecimal{Int8,2}(-0.01}`. Apologies.

NHDaly: hrmm sorry i don't remember.....

BUT importantly: so this will fail for BigInt, yeah?
```julia
julia> BigInt <: Signed
true

julia> typemin(BigInt)
ERROR: MethodError: no method matching typemin(::Type{BigInt})
```

Can you maybe add `&& hasmethod(typemin, (Type{T},))` into the if-block?

NHDaly: I guess we don't need it, but i do like it. It adds clarity to what it's doing. So I would say: keep it, and can you please add it to `div_with_overflow` as well?

NHDaly: aha, right! 👍 

Then we should make sure to test those cases too! Thanks for looking through that!

test/FixedDecimal.jl
```diff
@@ -789,6 +791,16 @@ end
...
         @test div_with_overflow(FD{Int16,1}(1639), FD{Int16,1}(0.5)) == (FD{Int16,1}(-3275.6), true)
 
+        @test rdiv_with_overflow(Int8(1), FD{Int8,2}(0.7)) == (FD{Int8,2}(-1.13), true)
+        @test rdiv_with_overflow(FD{Int16,2}(165), FD{Int16,2}(0.5)) == (FD{Int16,2}(-325.36), true)
+        @test rdiv_with_overflow(FD{Int16,2}(-165), FD{Int16,2}(0.5)) == (FD{Int16,2}(325.36), true)
+        @test rdiv_with_overflow(typemin(FD{Int64,8}), Int32(-1)) == (typemin(FD{Int64,8}), true)
+        @test rdiv_with_overflow(typemin(FD{Int64,0}), FD{Int64,0}(-1)) == (typemin(FD{Int64,0}), true)
+
+        @test fld_with_overflow(FD{Int8,2}(-1), FD{Int8,2}(0.9)) == (FD{Int8,2}(0.56), true)
+        @test fld_with_overflow(typemin(FD{Int64,0}), FD{Int64,0}(-1)) == (typemin(FD{Int64,0}), true)
```
NHDaly, line 801: per above, please add a case of non-0 `f`:
```
        @test fld_with_overflow(typemin(FD{Int64,2}), FD{Int64,2}(-1)) == (typemin(FD{Int64,2}), true)
```
or somethign like that

src/FixedPointDecimals.jl
mbarbar: 👍 done

mbarbar: >BUT importantly: so this will fail for BigInt, yeah?
Yep that's the problem :P

Have added the hasmethod for all introduced cases. More generally, we should do something like this (abstracted out) everywhere.

test/FixedDecimal.jl
mbarbar: Added this for rdiv_with_overflow an equivalent for fld (it doesn't overflow).

src/FixedPointDecimals.jl
NHDaly: Thanks. 👍 Yes, agreed.

test/FixedDecimal.jl
NHDaly: oops! Right, i meant `FD{Int64,2}(-0.01)` -- that is, i want to exercise the overflow case for a non-0 `f`.

I think it should be this:
```julia
julia> FixedPointDecimals.fld_with_overflow(typemin(FD{Int8,2}), FD{Int8,2}(-0.01))
(FixedDecimal{Int8,2}(-1.28), true)
```

test/FixedDecimal.jl
```diff
@@ -848,6 +875,17 @@ end
...
+            @test rdiv_with_overflow(Int8(1), FD{Int8,2}(0.8)) == (FD{Int8,2}(1.25), false)
+            @test rdiv_with_overflow(FD{Int64,8}(5), FD{Int64,8}(2)) == (FD{Int64,8}(2.5), false)
+            @test rdiv_with_overflow(FD{Int64,8}(5), FD{Int64,8}(0.5)) == (FD{Int64,8}(10), false)
+            @test rdiv_with_overflow(FD{Int128,0}(20000), Int32(5000)) == (FD{Int128,0}(4), false)
+
+            @test fld_with_overflow(typemax(FD{Int128,38}), FD{Int128,38}(1)) == (FD{Int128,38}(1), false)
+            @test fld_with_overflow(FD{Int64,8}(20.5), FD{Int64,8}(2.1)) == (FD{Int64,8}(9), false)
+            @test fld_with_overflow(FD{Int8,0}(-5), FD{Int8,0}(-1)) == (FD{Int8,0}(5), false)
+            @test fld_with_overflow(FD{Int8,2}(0.99), FD{Int8,2}(0.5)) == (FD{Int8,2}(1), false)
+            @test fld_with_overflow(typemin(FD{Int8,2}), FD{Int8,2}(-1)) == (FD{Int8,2}(1), false)
```
NHDaly, line 888:
```suggestion
            @test fld_with_overflow(typemin(FD{Int8,2}), FD{Int8,2}(-1)) == (FD{Int8,2}(1), false)
            @test FixedPointDecimals.fld_with_overflow(typemin(FD{Int8,2}), FD{Int8,2}(-0.01)) == (typemin(FD{Int8,2}), true)
```

test/FixedDecimal.jl
```diff
@@ -789,6 +791,17 @@ end
...
+        @test rdiv_with_overflow(Int8(1), FD{Int8,2}(0.7)) == (FD{Int8,2}(-1.13), true)
+        @test rdiv_with_overflow(FD{Int16,2}(165), FD{Int16,2}(0.5)) == (FD{Int16,2}(-325.36), true)
+        @test rdiv_with_overflow(FD{Int16,2}(-165), FD{Int16,2}(0.5)) == (FD{Int16,2}(325.36), true)
+        @test rdiv_with_overflow(typemin(FD{Int64,8}), Int32(-1)) == (typemin(FD{Int64,8}), true)
+        @test rdiv_with_overflow(typemin(FD{Int64,0}), FD{Int64,0}(-1)) == (typemin(FD{Int64,0}), true)
+        @test rdiv_with_overflow(typemin(FD{Int8,2}), FD{Int8,2}(-1)) == (typemin(FD{Int8,2}), true)
+
+        @test fld_with_overflow(FD{Int8,2}(-1), FD{Int8,2}(0.9)) == (FD{Int8,2}(0.56), true)
+        @test fld_with_overflow(typemin(FD{Int64,0}), FD{Int64,0}(-1)) == (typemin(FD{Int64,0}), true)
+        @test fld_with_overflow(FD{Int8,1}(7), FD{Int8,1}(0.5)) == (FD{Int8,1}(-11.6), true)
```
NHDaly, line 803:
```suggestion
        @test fld_with_overflow(FD{Int8,1}(7), FD{Int8,1}(0.5)) == (FD{Int8,1}(-11.6), true)
        @test FixedPointDecimals.fld_with_overflow(typemin(FD{Int8,2}), FD{Int8,2}(-0.01)) == (typemin(FD{Int8,2}), true)
```

test/FixedDecimal.jl
```diff
@@ -789,6 +791,17 @@ end
         @test div_with_overflow(typemin(FD{Int32,0}), FD{Int32,0}(-1)) == (typemin(FD{Int32,0}), true)
         @test div_with_overflow(FD{Int16,1}(1639), FD{Int16,1}(0.5)) == (FD{Int16,1}(-3275.6), true)
 
+        @test rdiv_with_overflow(Int8(1), FD{Int8,2}(0.7)) == (FD{Int8,2}(-1.13), true)
+        @test rdiv_with_overflow(FD{Int16,2}(165), FD{Int16,2}(0.5)) == (FD{Int16,2}(-325.36), true)
+        @test rdiv_with_overflow(FD{Int16,2}(-165), FD{Int16,2}(0.5)) == (FD{Int16,2}(325.36), true)
+        @test rdiv_with_overflow(typemin(FD{Int64,8}), Int32(-1)) == (typemin(FD{Int64,8}), true)
+        @test rdiv_with_overflow(typemin(FD{Int64,0}), FD{Int64,0}(-1)) == (typemin(FD{Int64,0}), true)
+        @test rdiv_with_overflow(typemin(FD{Int8,2}), FD{Int8,2}(-1)) == (typemin(FD{Int8,2}), true)
```
NHDaly, line 799:
```suggestion
        @test rdiv_with_overflow(typemin(FD{Int8,2}), FD{Int8,2}(-1)) == (typemin(FD{Int8,2}), true)
        @test rdiv_with_overflow(typemin(FD{Int8,2}), FD{Int8,2}(-0.01)) == (FD{Int8,2}(0), true)
```

test/FixedDecimal.jl
NHDaly: I've added the two test cases in PR review comment suggestions :)

NHDaly (approved):
LGTM after the final changes are applied. I'm apply them and merge

test/FixedDecimal.jl
mbarbar: Whoops :l thanks

PR #110, opened 2025-02-06
Better handle different backing types

mbarbar:
`trunc_int` is for primitive types only. `typemin` on unsigned types is 0, which you can obviously divide by -1.


mbarbar:
Closed because I branched off the wrong place :l

Issue #119, opened 2025-06-03
Julia v1.12 warning: type "Bool" was extended without explicit qualification or import.


inkydragon:
Just add this Pkg in Julia v1.12:
```jl
┌ FixedPointDecimals
│  WARNING: Constructor for type "Bool" was extended in `FixedPointDecimals` without explicit qualification or import.
│    NOTE: Assumed "Bool" refers to `Base.Bool`. This behavior is deprecated and may differ in future versions.`
│    NOTE: This behavior may have differed in Julia versions prior to 1.12.
│    Hint: If you intended to create a new generic function of the same name, use `function Bool end`.
│    Hint: To silence the warning, qualify `Bool` as `Base.Bool` in the method signature or explicitly `import Base: Bool`.
└
```
https://github.com/JuliaMath/FixedPointDecimals.jl/blob/f45fb681759049ad6cb6535900cfa55a0f625d96/src/FixedPointDecimals.jl#L729-L730

PR #121, opened 2025-07-16
Explicitly qualify Base.Bool constructor

rossviljoen:
Fixes a precompilation warning in julia 1.12:
```
┌ FixedPointDecimals
│  WARNING: Constructor for type "Bool" was extended in `FixedPointDecimals` without explicit qualification or import.
│    NOTE: Assumed "Bool" refers to `Base.Bool`. This behavior is deprecated and may differ in future versions.`
│    NOTE: This behavior may have differed in Julia versions prior to 1.12.
│    Hint: If you intended to create a new generic function of the same name, use `function Bool end`.
│    Hint: To silence the warning, qualify `Bool` as `Base.Bool` in the method signature or explicitly `import Base: Bool`.
└  
```

### DIFF
--- a/src/FixedPointDecimals.jl
+++ b/src/FixedPointDecimals.jl
@@ -727,7 +727,7 @@ end
 
 (::Type{T})(x::FD) where {T<:Union{AbstractFloat,Integer,Rational}} = convert(T, x)
 # Need to avoid ambiguity:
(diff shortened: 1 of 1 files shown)


rossviljoen:
fixes #119

